PR #999, opened 2024-08-14
[BUILD-575] feat: Show open-in-browser button in card preview of flashcard selector

RisingOrange:
In the card preview add a **Copy Note ID** button when shown on the AnkiHub Webapp and a **Open in browser** button when shown on Anki Desktop.

This is the addon-side of the task. This is the web app PR: https://github.com/AnkiHubSoftware/ankihub/pull/2260

## Related issues
https://ankihub.atlassian.net/jira/software/c/projects/BUILD/boards/1?selectedIssue=BUILD-575

## Proposed changes
- Add `is_on_anki=True` query argument to flashcard selector url in `settings.py`
- ref: Move js/pycmd message handling code from `reviewer.py` to `js_message_handling.py`
  - The reason is that the `OPEN_BROWSER_PYCMD` is now used for the flashcard selector as well, and not just in the reviewer. So it's better to move it to a separate module. Other pycmd code which could also be useful in multiple contexts was also moved to `js_message_handling.py`.

## How to reproduce
- Make sure that you have the `show_flashcards_selector_button` feature flag active
- Set up an AnkiHub deck with embeddings (generated using the `generate_vector_embeddings_for_decks.py` script) and a tag tree (generated using the `save_deck_tags_tree.py` script).
- Install the deck via the add-on
- Change `ANKING_DECK_ID` in `settings.py` to the deck id of the deck
- Open Anki, there should be a **Select flashcards** button next to the deck
- Open the flashcard selector, open the text search tab, search for something, open the card preview
- There should be a **Open in Browser** button on the card preview
- When you click the button the card should be opened in the Anki browser
- Also check for other search types

### DIFF
--- a/ankihub/entry_point.py
+++ b/ankihub/entry_point.py
@@ -10,7 +10,7 @@ from aqt.gui_hooks import profile_did_open, profile_will_close
 from . import LOGGER
 from .db import ankihub_db
 from .feature_flags import setup_feature_flags_in_background
-from .gui import browser, deckbrowser, editor, progress, reviewer
+from .gui import browser, deckbrowser, editor, js_message_handling, progress, reviewer
 from .gui.addons import setup_addons
 from .gui.auto_sync import setup_auto_sync
 from .gui.config_dialog import setup_config_dialog_manager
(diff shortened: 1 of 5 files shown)


augdiebold (approved):
Worked well!

PR #1011, opened 2024-09-16
[BUILD-632] fix: Setup happening after first AnkiWeb sync

RisingOrange:
In Anki 24.06 `AnkiQt.maybe_auto_sync_on_open_close` is called before the `profile_did_open hook`. (Both are called in `AnkiQt.loadProfile`)
We need to call `_on_profile_did_open` (which runs the add-on's setup) before `maybe_auto_sync_on_open_close`.
This PR makes this happen.

There is an Anki PR for restoring the previous order of calls in future Anki versions: https://github.com/ankitects/anki/pull/3421

## Related issues
https://ankihub.atlassian.net/jira/software/c/projects/BUILD/boards/1?selectedIssue=BUILD-632

## Proposed changes
- Wrap `maybe_auto_sync_on_open_close` to call the setup code (`_on_profile_did_open()`) before the original `maybe_auto_sync_on_open_close` function.
  - `maybe_auto_sync_on_open_close` is called when opening and when closing a profile. The code makes sure to only call it when a profile is opened.
- Bump Anki version to 24.06.3


ankihub/entry_point.py
```diff
@@ -30,13 +34,19 @@
...
+    ANKI_VERSION_24_06_00,
     ankihub_db_path,
+    assign_id_to_profile_if_not_exists,
     config,
+    get_anki_profile_id,
     setup_logger,
     setup_profile_data_folder,
 )
 
+CALL_ON_PROFILE_DID_OPEN_ON_MAYBE_AUTO_SYNC = ANKI_INT_VERSION >= ANKI_VERSION_24_06_00
```
abdnh, line 48: if the Anki PR gets merged, we probably should limit this to 24.06.x to avoid profile_did_open being called twice.

RisingOrange: It shouldn't get called twice given the code. Still it's good idea to restrict this behavior to 24.06.xx.
[Only use this on Anki 24.06.xx](https://github.com/AnkiHubSoftware/ankihub_addon/pull/1011/commits/408133e6e7ea8381a8491e20f50ddcc4a6337688)

abdnh (approved):

RisingOrange:
@abdnh I added some error handling to make sure that potential exception ins `_on_profile_did_open` don't disrupt the calling Anki code.

PR #1022, opened 2024-10-28
WIP: Refactor review_data.py to include a new dataclass and function …

pedroven:
…for retrieving daily review data since the last sync

## Related issues

- [x] Closes #

## Proposed changes
Describe the big picture of your changes here to communicate to the reviewers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## How to reproduce
Explain like if someone who doesn't know this project is reviewing your changes and how they can replicate this.

## Screenshots and videos

Paste here any screenshots or videos related to your changes, if applicable.

## Further comments
If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/ankihub/main/review_data.py
+++ b/ankihub/main/review_data.py
@@ -1,5 +1,7 @@
 import uuid
-from datetime import datetime, timedelta
+from collections import defaultdict
+from dataclasses import dataclass
+from datetime import date, datetime, timedelta
 from typing import List, Optional, Tuple
 
 import aqt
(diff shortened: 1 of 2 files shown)


ankihub/main/review_data.py
```diff
@@ -91,3 +93,55 @@ def _get_first_and_last_review_datetime_for_ah_deck(
...
+    rows = aqt.mw.col.db.all(
+        """
+        SELECT r.id, r.ease, r.time
+        FROM revlog as r
+        JOIN cards as c ON r.cid = c.id
+        WHERE r.id BETWEEN ? AND ?
+        """,
+        timestamp_last_sync_ms,
+        timestamp_yesterday_ms,
+    )
```
RisingOrange, line 116: :wrench: Shouldn't we use `end_of_yesterday` (as in the suggested code below) if we want to get the data for the whole last day? Otherwise the cut-off time is different depending on the current time.

```suggestion
    """Filter revlog entries between the date of the last sync and the end of yesterday
    group by days, and compile the data."""
    end_of_yesterday = datetime.now().replace(
        hour=0, minute=0, second=0, microsecond=0
    ) - timedelta(microseconds=1)
    timestamp_last_sync_ms = int(datetime.timestamp(last_sync) * 1000)
    timestamp_end_of_yesterday = int(datetime.timestamp(end_of_yesterday) * 1000)

    rows = aqt.mw.col.db.all(
        """
        SELECT r.id, r.ease, r.time
        FROM revlog as r
        JOIN cards as c ON r.cid = c.id
        WHERE r.id BETWEEN ? AND ?
        """,
        timestamp_last_sync_ms,
        timestamp_end_of_yesterday,
    )
```

tests/addon/test_unit.py
```diff
@@ -2906,3 +2908,69 @@ def test_with_migrating_content_from_old_note_type(self, use_new_templates: bool
...
+
+        mocker.patch(
+            "ankihub.main.review_data.aqt.mw.col.db.all", return_value=mock_rows
+        )
+        mock_ms_timestamp_to_datetime = mocker.patch(
+            "ankihub.main.review_data._ms_timestamp_to_datetime"
+        )
+        mock_ms_timestamp_to_datetime.side_effect = lambda ts: datetime.fromtimestamp(
+            ts / 1000
+        )
```
RisingOrange, line 2932: :pick: We don't need to patch this. The test passes without the patch.
```suggestion
```

tests/addon/test_unit.py
```diff
@@ -2906,3 +2908,69 @@ def test_with_migrating_content_from_old_note_type(self, use_new_templates: bool
...
+
+        expected_data = [
+            DailyCardReviewSummaryData(
+                total_cards_studied=2,
+                total_time_reviewing=70,
+                total_cards_marked_as_again=1,
+                total_cards_marked_as_hard=1,
+                total_cards_marked_as_good=0,
+                total_cards_marked_as_easy=0,
+                review_session_date=(last_sync + timedelta(seconds=1)).date(),
```
RisingOrange, line 2944: :pick: It seems like we don't need to subtract the 1 second here?

tests/addon/test_unit.py
```diff
@@ -2906,3 +2908,69 @@ def test_with_migrating_content_from_old_note_type(self, use_new_templates: bool
...
+            ),
+            DailyCardReviewSummaryData(
+                total_cards_studied=2,
+                total_time_reviewing=110,
+                total_cards_marked_as_again=0,
+                total_cards_marked_as_hard=0,
+                total_cards_marked_as_good=1,
+                total_cards_marked_as_easy=1,
+                review_session_date=(yesterday - timedelta(seconds=1)).date(),
```
RisingOrange, line 2953: :pick:It seems like we don't need to subtract the 1 second here?  (Same as the last comment.)

RisingOrange (changes requested):
I think we need to make a change to the time range we use to to filer the review entries.
Otherwise the code looks good!

RisingOrange (approved):

PR #1025, opened 2024-10-29
 [BUILD-748] feat: Add function to send the daily summary review data after the sync

pedroven:
## Related issues
- [BUILD-748](https://ankihub.atlassian.net/browse/BUILD-748)

## Proposed changes
This PR makes these changes:
* Add the function `send_daily_review_summaries` which will send to the AnkiHub server the compiled data of the daily reviews
* Call this function in the background after the user sync
* Update the `feature_flags.py` module and its use to avoid the need to update the dataclass if a new feature flag is created on the server

## How to reproduce
1. Run the instance of the ankihub server and make sure that the feature flag `daily_card_review_summary` is enabled
2. Open a deck and review a card
3. Open the anki database and update the ID of the newly created `revlog` row to the value of a timestamp of yesterday's date
4.  Click on the "Sync with AnkiHub" option on the Ankihub menu
5. Go to a shell instance of the Ankihub server and check if a register of the `DailyCardReviewSummary` was created successfully

[BUILD-748]: https://ankihub.atlassian.net/browse/BUILD-748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/operations/ankihub_sync.py
+++ b/ankihub/gui/operations/ankihub_sync.py
@@ -1,5 +1,6 @@
 from concurrent.futures import Future
 from dataclasses import dataclass
+from datetime import datetime
 from functools import partial
 from typing import Callable, List, Optional
 
@@ -13,8 +14,9 @@ from aqt.sync import get_sync_status
 from ... import LOGGER
 from ...addon_ankihub_client import AddonAnkiHubClient as AnkiHubClient
 from ...ankihub_client import API_VERSION, Deck
+from ...feature_flags import feature_flags
 from ...main.deck_unsubscribtion import uninstall_deck
-from ...main.review_data import send_review_data
+from ...main.review_data import send_daily_review_summaries, send_review_data
 from ...main.utils import collection_schema
 from ...settings import config
 from ..deck_updater import ah_deck_updater, show_tooltip_about_last_deck_updates_results
@@ -130,6 +132,16 @@ def _on_sync_done(future: Future, on_done: Callable[[Future], None]) -> None:
         send_review_data, on_done=_on_send_review_data_done
     )
 
+    last_summary_sent_date = config.get_last_summary_sent_date()
+    if (
+        feature_flags.get("daily_card_review_summary", False)
+        and last_summary_sent_date
+        and last_summary_sent_date.date() < datetime.now().date()
+    ):
+        aqt.mw.taskman.run_in_background(
+            send_daily_review_summaries, on_done=_on_send_daily_review_summaries_done
+        )
+
     if config.schema_to_do_full_upload_for_once():
         # Sync with AnkiWeb to resolve the pending full upload immediately.
         # Otherwise, Anki's Sync button will be red, and clicking it will trigger a full upload.
@@ -163,6 +175,26 @@ def _on_send_review_data_done(future: Future) -> None:
     else:
         LOGGER.error(  # pragma: no cover
             "Failed to send review data.", exc_info=exception
+        )
+
+
+def _on_send_daily_review_summaries_done(future: Future) -> None:
+    exception = future.exception()
+    if not exception:
+        LOGGER.info("Daily review summaries sent successfully")
+        return
+
+    # CollectionNotOpen is raised by Anki when trying to access the collection when it is closed.
+    # This happens e.g. when the sync is triggered by the user closing Anki. Then the task for sending review data
+    # starts and tries to access the collection, but it is already closed. We can ignore this error.
+    if "CollectionNotOpen" in str(exception):  # pragma: no cover
+        LOGGER.warning(
+            "Failed to send review summaries because the collection is closed.",
+            exc_info=exception,
+        )
+    else:
+        LOGGER.error(  # pragma: no cover
+            "Failed to send review summaries data.", exc_info=exception
         )
 
 
--- a/ankihub/gui/operations/ankihub_sync.py
+++ b/ankihub/gui/operations/ankihub_sync.py
@@ -134,15 +134,17 @@ def _on_sync_done(future: Future, on_done: Callable[[Future], None]) -> None:
     last_summary_sent_date = config.get_last_summary_sent_date()
     if not last_summary_sent_date:
         last_summary_sent_date = (
-            datetime.now().replace(hour=0, minute=0, second=0, microsecond=0)
-            - timedelta(microseconds=1)
-        ) - timedelta(days=1)
+            (
+                datetime.now().replace(hour=0, minute=0, second=0, microsecond=0)
+                - timedelta(microseconds=1)
+            )
+            - timedelta(days=1)
+        ).date()
 
     feature_flags = config.get_feature_flags()
     if (
         feature_flags.get("daily_card_review_summary", False)
-        and last_summary_sent_date
-        and last_summary_sent_date.date() < datetime.now().date()
+        and last_summary_sent_date < datetime.now().date()
     ):
         aqt.mw.taskman.run_in_background(
             lambda: send_daily_review_summaries(last_summary_sent_date),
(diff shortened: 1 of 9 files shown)


ankihub/gui/operations/ankihub_sync.py
```diff
@@ -130,6 +131,24 @@ def _on_sync_done(future: Future, on_done: Callable[[Future], None]) -> None:
...
+    feature_flags = config.get_feature_flags()
+    if (
+        feature_flags.get("daily_card_review_summary", False)
+        and last_summary_sent_date
+        and last_summary_sent_date.date() < datetime.now().date()
+    ):
+        aqt.mw.taskman.run_in_background(
+            lambda: send_daily_review_summaries(last_summary_sent_date),
+            on_done=_on_send_daily_review_summaries_done,
+        )
```
RisingOrange, line 150: :pick: Can we extract this into a `_maybe_send_daily_review_summaries` function?

ankihub/gui/operations/ankihub_sync.py
```diff
@@ -130,6 +131,24 @@ def _on_sync_done(future: Future, on_done: Callable[[Future], None]) -> None:
...
+    if not last_summary_sent_date:
+        last_summary_sent_date = (
+            datetime.now().replace(hour=0, minute=0, second=0, microsecond=0)
+            - timedelta(microseconds=1)
+        ) - timedelta(days=1)
+
+    feature_flags = config.get_feature_flags()
+    if (
+        feature_flags.get("daily_card_review_summary", False)
+        and last_summary_sent_date
```
RisingOrange, line 144: :pick: We can remove this line, because `last_summary_sent_date` is always set to a not-none value by the code above.
```suggestion
```

ankihub/gui/operations/ankihub_sync.py
RisingOrange, line 1: :wrench: I noticed that `get_daily_review_summaries_since_last_sync` is taking a `datetime` object as the argument:
https://github.com/AnkiHubSoftware/ankihub_addon/blob/97ab6ee232783ed29d79c3d7207b204bbedd7e96/ankihub/main/review_data.py#L98-L100

We are also storing a `datetime` in the private config.
https://github.com/AnkiHubSoftware/ankihub_addon/blob/97ab6ee232783ed29d79c3d7207b204bbedd7e96/ankihub/settings.py#L417 

It would make more sense  to use a `date` object for both of them (and at related places in the code.)
We only care about for which dates the review summaries were sent, not at which time.

I think with the current code there is an issue where some of the reviews for a day are not sent, because we are filtering out reviews which happened on the day the summaries were sent, but before the summaries were sent.

RisingOrange (changes requested):
I noticed an issue with the logic for the last sync time. I think we should use `datetime.date` instead of `datetime.datetime` for the `last_summary_sent_date` value.

I liked the changes to the feature flags module. However, there is an issue in the case that fetching the feature flag fails.

ankihub/gui/operations/ankihub_sync.py
pedroven: why this function should be private?

tests/addon/test_unit.py
```diff
@@ -2967,3 +2985,50 @@ def test_get_daily_review_data_no_reviews(mocker, anki_session_with_addon_data):
...
+
+    send_daily_review_summaries(last_summary_sent_date)
+
+    mock_get_daily_review_summaries.assert_called_once_with(last_summary_sent_date)
+    mock_anki_hub_client.send_daily_card_review_summaries.assert_called_once_with(
+        [mock_summary]
+    )
+    mock_config.save_last_summary_sent_date.assert_called_once()
+    assert (
+        mock_config.save_last_summary_sent_date.call_args[0][0] <= datetime.now().date()
```
RisingOrange, line 3011: :pick: I think it's better to be more strict here with the assertion.
```suggestion
        mock_config.save_last_summary_sent_date.call_args[0][0] == datetime.now().date()
```

tests/addon/test_unit.py
```diff
@@ -2967,3 +2985,50 @@ def test_get_daily_review_data_no_reviews(mocker, anki_session_with_addon_data):
...
+    mock_get_daily_review_summaries.return_value = []
+    mock_anki_hub_client = MockAnkiHubClient.return_value
+
+    send_daily_review_summaries(last_summary_sent_date)
+
+    mock_get_daily_review_summaries.assert_called_once_with(last_summary_sent_date)
+    mock_anki_hub_client.send_daily_card_review_summaries.assert_not_called()
+    mock_config.save_last_summary_sent_date.assert_called_once()
+    assert (
+        mock_config.save_last_summary_sent_date.call_args[0][0] <= datetime.now().date()
```
RisingOrange, line 3033: :pick: Same as above.
```suggestion
        mock_config.save_last_summary_sent_date.call_args[0][0] == datetime.now().date()
```

ankihub/gui/operations/ankihub_sync.py
RisingOrange: Because the add-on code base generally follows the convention that if a function is not used outside the file, it should be marked as private/internal by using an underline at the beginning of the function name.

ankihub/main/review_data.py
```diff
@@ -146,3 +148,18 @@ def get_daily_review_summaries_since_last_sync(
             )
         )
     return daily_card_review_data
+
+
+def send_daily_review_summaries(last_summary_sent_date: date) -> None:
+    """Send daily review summaries to the server."""
+    daily_review_summaries = get_daily_review_summaries_since_last_sync(
```
RisingOrange, line 155: :seedling: If a user doesn't use AnkiHub for a long time and then starts using it again, there can be a lot of daily summaries to send. 
Each summary is sent using its own request, so if they had a pause of 150 days, there would potentially be 150 requests...
It would be good to have an endpoint which accepts a list of summaries.

RisingOrange (commented):
Looks good, but there is one complicated expression which should be simplified.

ankihub/gui/operations/ankihub_sync.py
pedroven: but in this case, `send_daily_review_summaries` is being used outside the module `review_data` so in this case the function name should be `maybe_send_daily_review_summaries` instead of `_maybe_send_daily_review_summaries`

ankihub/main/review_data.py
pedroven: Good point, I will create these tasks as a nice to have (feel free to also create these tasks)

ankihub/gui/operations/ankihub_sync.py
pedroven: Anyway, being extremely opinionative here, I don't like the pattern of the "maybe" functions because I consider this harder to understand the function's behavior and hence is harder to debug/maintain. Here is a nice post about it: https://blog.benwinding.com/maybe-functions/. Since you are the developer who spends more time maintaining the addon, I left with you the decision to use the "maybe" function.

pedroven:
@RisingOrange I don't know if you noticed but seems like some tests are flaky, sometimes the CI pipeline fails, and sometimes not.

ankihub/gui/operations/ankihub_sync.py
RisingOrange: I think it's fine to use maybe functions in some situations.

> Disadvantages: 
> - Caller has less control, they depend on validation hidden in “maybe function”
> - Hides complexity but doesn’t actually remove it
> - Things that depend on the output now need to handle the maybeness case

> - Things that depend on the output now need to handle the maybeness case

In this case we are not depending on the output of the maybe function (it always returns `None`), so this is not a problem.

RisingOrange (approved):

RisingOrange:
> @RisingOrange I don't know if you noticed but seems like some tests are flaky, sometimes the CI pipeline fails, and sometimes not.

Yes, this is indeed an issue. I haven't got around to fixing it yet

PR #1026, opened 2024-10-30
[BUILD-762] perf: Use set when filtering notes to upsert into AnkiHub DB

RisingOrange:
The filtering of the notes to upsert takes a very long time when there are many notes to filter out. 
I tested with 30k notes and it took more than a minute.
With this change it takes a couple of seconds at most.

## Related issues
https://ankihub.atlassian.net/jira/software/c/projects/BUILD/boards/1?selectedIssue=BUILD-762

## Proposed changes
- Use set instead of list when filtering notes to upsert into AnkiHub DB


heitorado (approved):
Awesome find. 💯

PR #1039, opened 2024-11-25
[BUILD-853] Make the split screen header a different webview

abdnh:
## Related issues

 https://ankihub.atlassian.net/browse/BUILD-853

## Proposed changes

This moves the split screen tabs and header to a separate webview so we don't have to inject content into MH pages.


ankihub/gui/reviewer.py
```diff
@@ -62,20 +64,30 @@ def _setup_webview(self):
...
         self.webview.setPage(CustomWebPage(profile, self.webview._onBridgeCmd))
         self.webview.setUrl(aqt.QUrl(self.urls_list[0]["url"]))
-
-        self.webview.hide()
+        self.header_webview = aqt.webview.AnkiWebView()
+        self.header_webview.setSizePolicy(
+            aqt.QSizePolicy(
+                aqt.QSizePolicy.Policy.Expanding, aqt.QSizePolicy.Policy.Fixed
+            )
+        )
```
RisingOrange, line 86: :wrench: There is something wrong with the height of the header webview. For me it looks like this: 


I added a `self.header_webview.setFixedHeight(1)` call and then it looked much better:


However, I also noticed that there is a margin between the two webviews, we should try to remove this. We could do this in a different PR if that's better.
It looks like there is also a margin on the left and the right side of the sidebar, which we might also want to remove.

ankihub/gui/web/mh_integration_tabs.html
RisingOrange, line 1: :wrench: I think we should rename this file to `sidebar_tabs.html`, since the sidebar will also contain the chatbot.

RisingOrange (changes requested):

ankihub/gui/reviewer.py
abdnh: > I added a self.header_webview.setFixedHeight(1) call and then it looked much better:

Feel free to add it to the appropriate place. I couldn't reproduce the issue on Windows.

RisingOrange: I added an `adjustSize` instead and this fixed it for me.

Unfortunately there is still space between the webviews. Also, it looks like the header webview is slightly less wide than the other webview.

PR #1043, opened 2024-11-29
[BUILD-861] Update reviewer button tooltips

abdnh:
## Related issues
 https://ankihub.atlassian.net/browse/BUILD-861


pedroven (changes requested):
We probably need to update the position of the tooltip since the current  one is far from the buttons
image

pedroven (approved):

PR #1049, opened 2024-12-13
[BUILD-897] Show B&B and FA buttons for the AnKing deck only

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/BUILD-897

## Proposed changes

This PR hides the B&B and FA buttons side panel buttons for non-AnKing decks.

## How to reproduce

1. Go to the AnKing deck and confirm the buttons are still displayed.
2. Go to a different deck and confirm the buttons are not displayed.

## Screenshots and videos

In the AnKing deck:
image

In other decks:
image


pedroven (changes requested):
I tested it and it's working, but I believe we should add a new condition on the line of `if not split_screen_webview_manager:` to check if the deck is the Anking this way, we will avoid creating a new instance and we will not face some issues like this one:
image

Note: I wasn't able to see the "notifications" element on the buttons in my environment

pedroven (approved):

PR #1055, opened 2024-12-18
[BUILD-917] Update chatbot upsell

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/BUILD-917

## Proposed changes

This changes the chatbot icon if the user is non-premium and removes upsell handling (which is now handled in the web app)

## How to reproduce

- Check in https://github.com/AnkiHubSoftware/ankihub/pull/2492 in the web app.
- Sign in to a non-premium account and go to a deck where the chatbot is enabled.
- Confirm the chatbot icon is changed and open the chatbot to see the new upsell message.

## Screenshots and videos

image
image
image
image


ankihub/gui/web/templates.py
```diff
@@ -47,10 +47,11 @@ def get_reviewer_buttons_js(theme: str, enabled_buttons: List[str]) -> str:
         {
             "THEME": theme,
             "ENABLED_BUTTONS": ",".join(enabled_buttons),
+            # TODO
+            "IS_PREMIUM": str(True),
```
abdnh, line 51: I assume we don't have an API endpoint to check if the user is premium, so wee need to add one?

cjasm: The endpoint already exists, but we don't send the plan data through it. We will need to include the premium info on it!

There is the `/users/me` which retrieves information about the user but we need to include the premium info.
but we could create a new function on the `AnkihubClient` similar to the `owned_deck_ids` but to store the is_premium info.

ankihub/gui/web/reviewer_buttons.js
```diff
@@ -33,7 +34,8 @@ class AnkiHubReviewerButtons {
             },
             {
                 name: "chatbot",
-                iconPath: "/_chatbot_icon.svg",
+                iconPath: this.isPremium ? "/_chatbot_icon.svg" : "/_chatbot_icon_sleeping.svg",
+                iconPathDarkTheme: this.isPremium ? null : "/_chatbot_icon_sleeping_dark_theme.svg",
```
cjasm, line 38: :question: why the icon is `null`? is the file missing? should we ask the designers for the dark version?

abdnh: No separate icon is required because there's logic in the file to change the foreground/background colors and that works for the this icon but not for the sleeping version (thus the separate _chatbot_icon_sleeping_dark_theme.svg version)

image

ankihub/ankihub_client/ankihub_client.py
```diff
@@ -1290,6 +1290,14 @@ def send_daily_card_review_summaries(
         if response.status_code != 201:
             raise AnkiHubHTTPError(response)
 
+    def is_premium_user(self) -> bool:
```
cjasm, line 1293: Could you add the `is_trialing` info, because the chatbot is allowed for users who are also on trial period.

ankihub/gui/web/reviewer_buttons.js
```diff
@@ -35,7 +36,8 @@ class AnkiHubReviewerButtons {
...
```
cjasm, line 40: Update the check to verify if the user is trialing

abdnh:
One issue with only checking premium & trialing status is that it doesn't account for other types of access. Admins for example have access to the feature but are neither premium nor trialing, so in this case, there will be a mismatch in the UI as the add-on will show the sleeping icon but the user will have access to the chatbot. Probably not a huge deal for admins, but I thought of pointing it out in case you think it can affect other types of users @cjasm

cjasm (dismissed):
LGTM

abdnh:
Still looking into the failing tests. I think I'll bite the bullet and set up a working testing environment today...

abdnh:
Fixed the tests finally!

cjasm (approved):

abdnh:
@cjasm Can you merge this? I can't merge it for some reason.

PR #1076, opened 2025-01-14
[BUILD-974] feat: Add config options for MH steps

RisingOrange:
We want to allow users to select for which USMLE step they want to see BB/FA resources in the sidebar.

## Related issues
https://ankihub.atlassian.net/jira/software/c/projects/BUILD/boards/1?selectedIssue=BUILD-914

## Proposed changes
- Add nested checkboxes to the config dialog
  - [Split up MH options into step 1 and step 2](https://github.com/AnkiHubSoftware/ankihub_addon/pull/1076/commits/99b0cf5b4c35e5f5fa0b5780ad3a9d5fbcd8ce1f)
- Filter out resources based on the values

## How to reproduce
- Change the options in the config dialog and check that only resources for the enabled USMLE steps are shown in the reviewer sidebar


tests/addon/test_unit.py
```diff
@@ -3055,62 +3059,80 @@ def test_send_daily_review_summaries_without_data(mocker):
     mock_anki_hub_client.send_daily_card_review_summaries.assert_not_called()
 
 
+def url_mh_integrations_preview(slug: str) -> str:
```
abdnh, line 3062: Can we use settings.url_mh_integrations_preview() instead of this?

ankihub/gui/config_dialog.py
```diff
@@ -66,3 +75,42 @@ def _general_tab(conf_window) -> None:
...
+    step_1_checkbox = container_inner.checkbox(
+        f"{key_prefix}_step_1", description="USMLE Step 1"
+    )
+    step_2_checkbox = container_inner.checkbox(
+        f"{key_prefix}_step_2", description="USMLE Step 2"
+    )
+
+    def update_main_checkbox() -> None:
+        main_checkbox.blockSignals(True)
+        main_checkbox.setChecked(
```
abdnh, line 104: Maybe we can set Qt.CheckState.PartiallyChecked if only one step is enabled?

abdnh (dismissed):

tests/addon/test_unit.py
RisingOrange: No, because the `config.app_url` is not initialized when the parametrization data is initialized. I'll add a comment

ankihub/gui/config_dialog.py
RisingOrange: Good idea!

abdnh (approved):

PR #1085, opened 2025-01-23
fix: show dialog titles on macOS and force a consistent button layout

abdnh:
## Related issues

This addresses issues reported by design team in https://ankihub.atlassian.net/browse/BUILD-921

## Proposed changes

- Qt doesn't show dialog titles on macOS by default (following Apple's guidelines). In BUILD-921, we convey important info in the title, so we need to show it somehow. The solution used here is to add the title to the main text body on macOS, with a larger bold font.
- Another issue reported is that the button layout is different on macOS. I solved this by simply telling Qt to follow the "standard" button layout used on Windows.

## How to reproduce

- Try to trigger any code path that uses show_dialog() on macOS and notice the change.

## Screenshots and videos

Untitled


cjasm (approved):
LGTM

abdnh:
@cjasm Can you merge it? Can't merge it myself 😐

PR #1089, opened 2025-01-29
[BUILD-1016] fix: Hide ankihub_id field in editor

RisingOrange:
The `ankihub_id` field should be hidden in the editor, but it isn't, due to some change to Anki.

**With this PR**

## Related issues
https://ankihub.atlassian.net/jira/software/c/projects/BUILD/boards/1?selectedIssue=BUILD-1016
https://forums.ankiweb.net/t/anki-24-11-addon-using-svelte-failing-with-cannot-require-svelte-internal/53068/6

## Proposed changes
- Use `setTimeout` instead of `require("svelte/internal").tick().then`


abdnh:
I remember I had to set a timeout (usually 100 or even 1000 in some add-on) but that's using document.querySelectorAll() to get the fields rather than `require('anki/NoteEditor')`. This does appear to work reliably.

abdnh (approved):

PR #1094, opened 2025-02-12
AI editor

andrewsanchez:


ankihub/labs/llm/llm.py
```diff
@@ -0,0 +1,334 @@
...
+    """Install llm and additional providers using uv if not already installed."""
+    # TODO Prompt users to set up their API keys.
+    try:
+        subprocess.run(["llm", "--version"], capture_output=True, check=True)
+        print("llm is already installed")
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        try:
+            # Install base llm package
+            subprocess.run(
+                ["uv", "install", "llm"],
```
RisingOrange, line 54: Just using `uv install llm` doesn't work
```suggestion
                ["uv", "pip", "install", "llm"],
```

ankihub/entry_point.py
```diff
@@ -238,6 +239,10 @@ def _general_setup():
         "Set up feature flag fetching (flags will be fetched in the background)."
     )
 
+    if config.labs_enabled:
+        llm.setup()
+        LOGGER.info("Set up LLM prompt functionality.")
+
```
andrewsanchez, line 245: @RisingOrange One possibility to implement "AnkiHub Labs": if labs is enabled, we clone/pull from a separate mono repo on GitHub that contains a bunch of community contributed functionality. Users can individually toggle them on/off. We could accept or reject PRs to this mono repo that are basically a bunch of mini standalone add-ons. Except, we could provide facilities for quick and easy development, web capabilities, and maybe mobile integrations. We could provide a replit template, cursor rules, and copilot prompts to make developing these really easy with AI.

ankihub/labs/llm/llm.py
```diff
@@ -0,0 +1,482 @@
...
+                    "curl -LsSf https://astral.sh/uv/install.sh | sh",
+                    shell=True,
+                    check=True,
+                )
+            elif platform.system() == "Windows":
+                subprocess.run(
+                    'powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"',
+                    shell=True,
+                    check=True,
+                )
```
RisingOrange, line 182: :wrench: To make it support Linux:
```suggestion
            if platform.system() == "Windows":
                subprocess.run(
                    'powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"',
                    shell=True,
                    check=True,
                )
            else:  # macOS and Linux
                subprocess.run(
                    "curl -LsSf https://astral.sh/uv/install.sh | sh",
                    shell=True,
                    check=True,
                )
```

ankihub/labs/llm/llm.py
```diff
@@ -0,0 +1,482 @@
...
+                    showWarning(f"Failed to install {provider}: {str(e)}")
+
+        except subprocess.CalledProcessError as e:
+            showWarning(f"Failed to install llm: {str(e)}")
+
+
+def setup() -> None:
+    """Set up the LLM prompt functionality."""
+    _check_and_install_uv()
+    _install_llm()
```
RisingOrange, line 224: :wrench: These operations should be done in the background to not block Anki's UI during the installation

RisingOrange:
Some UI elements are unreadable on dark mode:

RisingOrange:
@andrewsanchez Using `llm` require an API key for the LLM provider being set up on the user's device. Will we require users to provide an API key?

ankihub/labs/llm/llm.py
```diff
@@ -0,0 +1,481 @@
...
+        try:
+            # Install base llm package
+            subprocess.run(
+                ["uv", "tool", "install", "llm"],
+                check=True,
+                capture_output=True,
+            )
+            tooltip("Successfully installed llm")
+
+            # Install additional providers
```
RisingOrange, line 205: :wrench: We should install the providers even if the user already has `llm` installed

andrewsanchez:
> @andrewsanchez Using `llm` requires an API key for the LLM provider being set up on the user's device. Will we require users to provide an API key?

This PR is just a quick proof of concept in order to test out some workflows internally with a few people. So I will provide API keys that testers on our team can use. If we decide to put this in the hands of real users, we would need to migrate the `llm` related code to our infrastructure or create a new service/API. At that point, we could consider continuing to use `llm` or use langchain or something else instead.

ankihub/labs/llm/llm.py
andrewsanchez: What's the recommended way to do that? This is cursor's suggestion:

```python
    from aqt.operations import QueryOp
    from aqt.utils import tr

    def _background_setup():
        _check_and_install_uv()
        _install_llm()

    op = QueryOp(
        parent=active_window_or_mw(),
        op=lambda _: _background_setup(),
        success=lambda _: None,
    )
    op.with_progress(tr.actions_processing()).run_in_background()
```

RisingOrange: Added a commit with similar code: https://github.com/AnkiHubSoftware/ankihub_addon/commit/98e2ba9ae78cf44ef1a44656e2ec242c9bd83996

ankihub/labs/llm/llm.py
```diff
@@ -272,6 +277,30 @@ def _on_execute(self) -> None:
...
+            local_template_path = (
+                TemplateManager._local_templates_dir / f"{self.template_name}.yaml"
+            )
+            self.template_content = local_template_path.read_text(encoding="utf-8")
+            self.yaml_data = yaml.safe_load(self.template_content)
+            if not isinstance(self.yaml_data, dict):
+                self.yaml_data = {}
+            self._update_editors()  # Update the editors with the reset content
+            TemplateManager.save_template(self.template_name, self.template_content)
+            tooltip("Template reset.", parent=self)
```
andrewsanchez, line 302: Not sure why but this results in some funky yaml, @RisingOrange :

```yaml
model: gpt-4o
system: "You are an expert at creating effective flashcards. Analyze the provided\
  \ cloze deletion note and respond with an improved note according to the users instructions.\n\
  Follow these guidelines:\n- Use cloze deletions to hide key information.\n- Use\
  \ the {{c1::hidden text}} format for cloze deletions. You can also use {{c2::hidden\
  \ text}}, {{c3::hidden text}}, and so on. This will result in multiple cards being\
  \ created for the note by Anki. Cloze deletions with the same number will be hidden\
  \ on the same card, while all the other cloze deletions won't be used for this card.\n\
  - Most notes should only have one cloze deletion. Focus on the highest yield concepts\
  \ and facts. \"high-yield\" means the information is most important or most frequently\
  \ tested on exams.\n- Create focused cards which tests a single concept or fact.\n\
  - Ensure the remaining context provides enough information for the user to recall\
  \ the hidden content.\n\n## Common Mistakes\nAvoid these common mistakes when creating\
  \ cloze deletion cards:\n- Occluding too much information, making the card impossible\
  \ to answer.\n  - E.g., \"The {{c1::sky}} is {{c1::blue}}.\" This card is impossible\
  \ to answer because both \"sky\" and \"blue\" are occluded.\n\n\nHere are some examples\
  \ of well-constructed cloze deletion notes:\n- Persistent {{c1::infection}} is the\
  \ most common symptom of chronic sinopulmonary disease\n- Chronic sinopulmonary\
  \ disease commonly presents with chronic {{c1::cough}} and {{c1::sputum}} production.\n\
  - 11-deoxycorticosterone (DOC) may be converted to {{c1::corticosterone}} via the\
  \ enzyme {{c2::11\u03B2-hydroxylase}}.\n- 11\u03B2-hydroxylase deficiency presents\
  \ with {{c1::hypokalemia}} and {{c1::hypertension}}.\n- The {{c1::orbit}} is the\
  \ bony socket housing the eyeball and contains muscles for {{c2::eye movement}}\
  \ and {{c2::eyelid opening}}.\n- The upper portion of the nasal septum is formed\
  \ by the {{c1::ethmoid}} bone; the lower portion is formed by the {{c2::vomer}}\
  \ bone.\n- {{c1::Bone marrow}} within bones stores {{c2::fat}} and produces {{c3::blood\
  \ cells}}.\n- Two main side effects of osmotic laxatives include {{c1::diarrhea}}\
  \ and {{c1::dehydration}}.\n- Newly synthesized cGMP (from guanylyl cyclase activity)\
  \ activates {{c1::protein kinase G}}, which phosphorylates specific proteins responsible\
  \ for physiologic actions.\n- The posterior lobe of the pituitary gland is derived\
  \ from {{c1::neural}} tissue.\n- The skeleton protects {{c1::internal organs}} and\
  \ stores minerals like {{c2::calcium}} and {{c2::phosphate}}.\n\nImprove the provided\
  \ cloze deletion note based on the instructions above. Ensure that the card is clear,\
  \ and effective for learning and recall. Respond only with the improved note. Do\
  \ not provide any pre-amble, explanation, or surrounding xml tags.\n"
prompt: 'Here is the card to improve:

  <original note>

  $input

  </original note>

  '
```

andrewsanchez: actually not sure if it's this or saving user changes. I'll take a look

andrewsanchez: ya I think it happens when users try to save the template changes but not sure why yet.

andrewsanchez:
https://simonwillison.net/2025/Mar/6/aider-using-uv-as-an-installer/

PR #1099, opened 2025-02-18
feat: Handle note type changes which require a full AnkiWeb sync

RisingOrange:
Some changes related to note types trigger a full sync with AnkiWeb on the next AnkiWeb sync. 

Full syncs can cause un-synced data from other devices to get lost. We should prevent that.

We have some code for handling changes that require full syncs, but it's currently only used if the change happens during a deck's installation, not during normal sync.

This task adds handling for changes which will require full sync that occur during syncing with AnkiHub.

Figma: https://www.figma.com/design/rWzQ7PPqAXf1piz5yTntaT/%F0%9F%91%91--Ankihub?node-id=22105-197682&t=bwKL57ihVmh9126K-0

## Related issues
https://ankihub.atlassian.net/jira/software/c/projects/BUILD/boards/1?selectedIssue=BUILD-1041

## Proposed changes
The `AnkiHubImporter` raises `ChangesRequireFullSyncError` before making changes that would require a full sync, unless `raise_if_full_sync_required=False` is passed to `AnkiHubImporter.import_ankihub_deck`.
The changes which require a full sync are:
- Adding/removing fields of a note type
- Adding/removing card templates of a note type
- Changing note types of notes

The importer checks for these changes by comparing the note types which are about to be imported to the note types in the Anki database.

During the sync (in `ankihub.gui.operations.ankihub_sync`), we handle the `ChangesRequireFullSyncError` by showing the `ChangesRequireFullSyncDialog`:
- When the user clicks **Run Full Sync**, we run `ah_deck_updater.update_decks_and_media again`, passing `raise_if_full_sync_required=False`, then do a full upload to AnkiWeb after the AnkiHub sync.
- When the user clicks **Skip for now**, we skip the AnkiHub sync

### Syncing changes that require full sync
- Install an AnkiHub deck
- Log in to AnkiWeb and sync
- Remove a field from a note type of the deck (Tools -> Note Types -> Select the note type -> Fields)
- Sync with AnkiHub
- `ChangesRequireFullSyncDialog` should show up
- The note type should be listed in the **Note type updates** section once you expand it
- Click the **Skip** button to skip the sync
- Sync with AnkiHub again
- `ChangesRequireFullSyncDialog` should show up again
- Click the "Run Full Sync" button
- Check that the removed field exists again on the note type

### Script to show the dialog
You can run this in Anki's debug console to show the `ChangesRequireFullSyncDialog`:
```python
import aqt
from ankihub.gui.changes_require_full_sync_dialog import ChangesRequireFullSyncDialog, ChangesRequireFullSyncError

mid = aqt.mw.col.models.all_names_and_ids()[0].id
error = ChangesRequireFullSyncError(set([mid]))
dialog = ChangesRequireFullSyncDialog(error, parent=aqt.mw)
dialog.show()
```


ankihub/main/utils.py
```diff
@@ -442,21 +453,26 @@ def note_type_with_updated_templates_and_css(
     """
 
     updated_templates = []
-    for template_idx, old_template in enumerate(old_note_type["tmpls"]):
```
RisingOrange, line 445: `note_type_with_updated_templates_and_css` was updated to support updating a note type when templates were added or removed.

ankihub/main/importing.py
```diff
@@ -187,7 +190,106 @@ def import_ankihub_deck(
...
+    ) -> None:
+        def field_tuples(flds: List[Dict]) -> List[Tuple[int, str]]:
+            return [(field["ord"], field["name"]) for field in flds]
+
+        note_types_with_field_conflicts: List[Tuple[NotetypeDict, NotetypeDict]] = []
+        for mid, remote_note_type in remote_note_types.items():
+            local_note_type = aqt.mw.col.models.get(mid)
+
+            # TODO Allow extra local fields
+            if not field_tuples(local_note_type["flds"]) == field_tuples(
```
abdnh, line 219: This got me thinking: we probably should change `add_note_type_fields()` to avoid changing ordinals of existing fields if we don't support that.

ankihub/main/utils.py
```diff
@@ -224,7 +225,9 @@ def get_note_types_in_deck(did: DeckId) -> List[NotetypeId]:
     return aqt.mw.col.db.list(query)
 
 
-def change_note_types_of_notes(nid_mid_pairs: List[Tuple[NoteId, NotetypeId]]) -> None:
+def change_note_types_of_notes(
+    nid_mid_pairs: List[Tuple[NoteId, NotetypeId]], raise_if_changes_required=False
```
abdnh, line 229: Nitpick: rename to `raise_if_full_sync_required`.

ankihub/main/importing.py
```diff
@@ -187,7 +190,106 @@ def import_ankihub_deck(
...
+                    and projekt_anki_note_types_addon_installed
+                )
+            )
+
+            if (
+                self._raise_if_full_sync_required
+                and use_new_templates_and_css
+                and len(local_note_type["tmpls"]) != len(remote_note_type["tmpls"])
+            ):
+                raise ChangesRequireFullSyncError(
```
abdnh, line 283: Nitpick: can we move this outside the loop so all types with differing templates are reported all at once (If I'm reading this correctly)?

ankihub/main/reset_local_changes.py
```diff
@@ -46,6 +46,8 @@ def reset_local_changes_to_notes(
         subdecks_for_new_notes_only=deck_config.subdecks_enabled,
         suspend_new_cards_of_new_notes=deck_config.suspend_new_cards_of_new_notes,
         suspend_new_cards_of_existing_notes=deck_config.suspend_new_cards_of_existing_notes,
+        # TODO Warn user if full sync is required when resetting local changes
+        raise_if_full_sync_required=False,
```
abdnh, line 50: Do you plan to tackle this in this PR?

abdnh (approved):

ankihub/main/importing.py
RisingOrange: Yes, I moved it out of the loop now. 

> so all types with differing templates are reported all at once

(Previously only one of the note types with differing templates would be reported, but once the user runs the full sync all of them would be updated.)

ankihub/main/importing.py
RisingOrange: I think we could actually remove the need for the ordinals to match and just require the remote field names to exist locally.
Maybe we could do that in https://ankihub.atlassian.net/jira/software/c/projects/BUILD/boards/1?selectedIssue=BUILD-1059?

About `add_note_type_fields`, it seems to me like there could be an issue with the ordinals, because the function first updates the ordinals for all fields based on the original note type, and then it later adjusts the special field’s ordinal by moving it to the end.

ankihub/main/reset_local_changes.py
RisingOrange: No, I think this warning is not that important. I'll look into it and perhaps add this in a different PR

abdnh (approved):

PR #1090, opened 2025-02-05
[BUILD-1013] Allow deck owners to add and change note types

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/BUILD-1013

## TODO

- [x] API
- [ ] ~fetch and create new notetypes when syncing~

## Proposed changes
Describe the big picture of your changes here to communicate to the reviewers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## How to reproduce
Explain like if someone who doesn't know this project is reviewing your changes and how they can replicate this.

### Deck Management buttons

image

### Add Note Type flow

image
image

### Add Field flow

image
image

### Update Note Styling flow

image

## Further comments
If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

PR #1101, opened 2025-02-19
[BUILD-1020] Implement 'Add Field' button

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/BUILD-1020

## Proposed changes

This implements the logic part of the 'Add Field' button added in https://github.com/AnkiHubSoftware/ankihub_addon/pull/1090

## How to reproduce

- Add a new field to an AnkiHub note type from _Tools > Manage Note Types > Fields > Add_.
- Go to _AnkiHub > Deck Management > Publish Field_ and choose the relevant note type and field you want to upload.
- Log in to a different account subscribed to the deck and confirm the new field is added after syncing.

## TODO

- [x] tests


ankihub/main/note_type_management.py
```diff
@@ -33,8 +34,41 @@ def add_note_type(ah_did: uuid.UUID, note_type: NotetypeDict) -> NotetypeDict:
...
+        ),
+        None,
+    )
+    if ankihub_id_field_idx is not None:
+        db_note_type["flds"][ankihub_id_field_idx]["ord"] = (
+            len(db_note_type["flds"]) - 1
+        )
+        ankihub_id_field = db_note_type["flds"].pop(ankihub_id_field_idx)
+        db_note_type["flds"].append(ankihub_id_field)
+    db_note_type = client.update_note_type(ah_did, db_note_type)
```
cjasm, line 68: :wrench: I don't think we need to send all the note type data. That's the reason for using a PATCH
```suggestion
    payload = { 'fields': db_note_type.get("flds") }
    db_note_type = client.update_note_type(ah_did, payload)
```

ankihub/main/note_type_management.py
```diff
@@ -33,8 +34,41 @@ def add_note_type(ah_did: uuid.UUID, note_type: NotetypeDict) -> NotetypeDict:
...
 
 
-def update_note_type_fields(note_type: NotetypeDict, fields: List[str]) -> None:
-    print("update_note_type_fields", note_type["name"], fields)
+def add_note_type_fields(
+    ah_did: uuid.UUID, note_type: NotetypeDict, new_field_names: List[str]
+) -> NotetypeDict:
+    client = AddonAnkiHubClient()
+
+    db_note_type = ankihub_db.note_type_dict(ah_did, note_type["id"])
```
cjasm, line 42: It is pythonic to use `.get()` when accessing a variable on a dict and `dict[key]` when attributing a value. `dict[key]` may raise `KeyError`

```suggestion
    db_note_type = ankihub_db.note_type_dict(ah_did, note_type.get("id"))
```

ankihub/main/note_type_management.py
```diff
@@ -33,8 +34,41 @@ def add_note_type(ah_did: uuid.UUID, note_type: NotetypeDict) -> NotetypeDict:
...
+            idx
+            for idx, field in enumerate(db_note_type["flds"])
+            if field["name"] == ANKIHUB_NOTE_TYPE_FIELD_NAME
+        ),
+        None,
+    )
+    if ankihub_id_field_idx is not None:
+        db_note_type["flds"][ankihub_id_field_idx]["ord"] = (
+            len(db_note_type["flds"]) - 1
+        )
```
cjasm, line 65: :question: Would this conflict with Anki note type's field order?

I mean, locally, I'll have front: 0, middle: 1, back: 2. However, on the server we will have front:0 back:1.
What will happen?

ankihub/main/note_type_management.py
abdnh: `id` is guaranteed to exist here. It's also probably better to fail with `KeyError` rather than potentially mask a bug if `ankihub_db.note_type_dict()` is OK with `note_type_id=None`.

ankihub/main/note_type_management.py
abdnh: Hmm, I think that should be fine as long as we maintain relative order of fields. The deck owner's local field ordinals can be different anyway because we allow choosing which fields to upload.

cjasm (approved):

PR #1104, opened 2025-02-24
fix: UI fixes for note type creation

abdnh:
## Proposed changes

Fix more issues reported during QA of [BUILD-1013](https://ankihub.atlassian.net/browse/BUILD-1013):

- Update titles of confirmation dialogs.
- Revert to using default QDialogButtonBox button layout
-  Report note type creation errors (e.g. "A note type with this name already exists.")

[BUILD-1013]: https://ankihub.atlassian.net/browse/BUILD-1013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/decks_dialog.py
+++ b/ankihub/gui/decks_dialog.py
@@ -648,7 +648,7 @@ class DeckManagementDialog(QDialog):
             confirm = ask_user(
                 "<b>Proceed?</b><br><br>"
                 "Confirm to publish this note type to all AnkiHub users of your deck.<br><br>",
-                title="Publish note type to all users",
+                title="Publish note type",
                 no_button_label="Cancel",
             )
             if not confirm:
@@ -704,7 +704,7 @@ class DeckManagementDialog(QDialog):
                     "<b>Proceed?</b><br><br>"
                     "Confirm to publish the fields to all AnkiHub users of your deck.<br><br>"
                     + NOTE_TYPE_CHANGES_WARNING,
-                    title="Publish fields to all users",
+                    title="Publish fields",
                     no_button_label="Cancel",
                 )
                 if not confirm:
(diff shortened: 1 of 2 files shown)


ankihub/gui/decks_dialog.py
```diff
@@ -653,9 +655,17 @@ def on_note_type_selected(ret: SearchableSelectionDialog) -> None:
...
+                add_note_type(self._selected_ah_did(), note_type)
+                tooltip("Note type published", parent=aqt.mw)
+            except AnkiHubHTTPError as e:
+                errors = "\n".join(
+                    itertools.chain.from_iterable(e.response.json().values())
+                )
+                if e.response.status_code == 400:
+                    showWarning(errors)
+                else:
+                    raise e
```
RisingOrange, line 668: :wrench: We already have code for displaying error messages returned by the API in the error handler:
https://github.com/AnkiHubSoftware/ankihub_addon/blob/97a39e5dd404ef70728b15813b36e85e89f3199e/ankihub/gui/errors.py#L383-L387
  
It expects the error messages to be stored under the `detail` key.

We are using this for example for example to display the "You already have a deck with that name." message. It gets returned by the API here:
https://github.com/AnkiHubSoftware/ankihub/blob/d30a9fe76e4743969f784459a01332f2c08eb97e/ankihub/decks/api/views.py#L112-L115

So, for consistency and conciseness it might be better to change the API to return the "A note type with this name already exists." message under the `detail` key. 

Or we could adjust the code in the error handler to show messages even if there is no `detail` key:
```python
    error_message = "\n".join(
        itertools.chain.from_iterable(response_data.values())
    )
    if error_message:
        LOGGER.info("AnkiHubRequestError was handled", error_message=error_message)
        show_error_dialog(error_message, title="Oh no!")
        return True

    return False
```

RisingOrange: > change the API to return the "A note type with this name already exists." message under the detail key.

This option is probably better because this way we will still have a way to distinguish between `AnkiHubHTTPError` objects for which the error message should be shown to the user and those where the `ErrorDialog` should be shown and the error reported

abdnh: OK, created a JIRA task for this: https://ankihub.atlassian.net/browse/BUILD-1071

PR #1108, opened 2025-02-25
fix: Disable note type buttons if not applicable

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/BUILD-1013

## Proposed changes

This updates the UI to disable the buttons if they are not applicable, according to the Figma updates.

## How to reproduce

1. Go to the Deck Management screen.
2. Select a deck you own and notice the state of the buttons.
3. The "Publish note type" button should be disabled if you have no note types to upload to AnkiHub. This is unlikely because most of the time you'll have Anki's standard types there. But you can confirm this is working by deleting all non-AnkiHub types from _Tools > Mange Note Types_.
4. The "Publish field" button should be disabled if there are no new fields in any types associated with the deck. This button will probably show as disabled most of the time. You can confirm it's working by adding a new field to any relevant AnkiHub note type from _Tools > Manage Note Types_.
5. The "Publish style/template updates" button should be disabled if you made no changes to templates/styling sections of the note types. You can confirm it's working by making changes to the templates from _Tools > Manage Note Types > Cards_ (make sure your changes are above the ANKIHUB_END comment).

## Screenshots and videos

image

### DIFF
--- a/ankihub/gui/decks_dialog.py
+++ b/ankihub/gui/decks_dialog.py
@@ -35,6 +35,7 @@ from ..main.deck_unsubscribtion import unsubscribe_from_deck_and_uninstall
 from ..main.note_type_management import (
     add_note_type,
     add_note_type_fields,
+    new_fields_for_note_type,
     note_types_with_template_changes_for_deck,
     update_deck_templates,
 )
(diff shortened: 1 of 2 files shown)


ankihub/gui/decks_dialog.py
```diff
@@ -589,24 +590,69 @@ def _setup_box_note_types(
...
+        mids_with_updates = note_types_with_template_changes_for_deck(
+            self._selected_ah_did()
+        )
+        return [
+            n.name
+            for n in self._get_note_type_names_and_ids_for_deck(
+                self._selected_ah_did(), assigned_to_deck=True
+            )
+            if n.id in mids_with_updates
+        ]
```
ddevdan, line 651: 🔧 This function is returning True after the update is complete, because the note_type has "\n\n" at the end. 

image

Because of it the state of the button is not updated.

https://github.com/user-attachments/assets/7bf8b391-b973-4ae4-9406-977cc8438a84

ddevdan (changes requested):

ankihub/gui/decks_dialog.py
abdnh: This is probably just the result of the backend modifying `tmpls` and `css` (probably sanitization). `update_note_type_templates_and_styles()` assumes the backend does no further modifications of its own so we're not updating the local note type based on the one returned by `client.update_note_type()`.

ddevdan (approved):

PR #1112, opened 2025-03-04
fix: Exclude note types with duplicate names from publication flow

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/BUILD-1013

## Proposed changes

Because we clone chosen note types under new names, the old ones will still be shown next time, but we show an error message when they are chosen. This PR excludes any note types with duplicate names to reduce confusion.

## How to reproduce

- Publish any new note type.
- Click "Publish field" again and notice the previously chosen note type is no longer shown.

### DIFF
--- a/ankihub/gui/operations/username.py
+++ b/ankihub/gui/operations/username.py
@@ -14,7 +14,7 @@ def _fetch_username_in_background() -> None:
         try:
             username = client.get_user_details()["username"]
         except (AnkiHubRequestException, AnkiHubHTTPError) as exc:
-            LOGGER.error(f"Failed to fetch username: {exc}")
+            LOGGER.warning(f"Failed to fetch username: {exc}")
     config.save_username(username)
 
 
--- a/ankihub/gui/operations/username.py
+++ b/ankihub/gui/operations/username.py
@@ -1,0 +1,29 @@
+import aqt
+
+from ... import LOGGER
+from ...addon_ankihub_client import AddonAnkiHubClient as AnkiHubClient
+from ...ankihub_client import AnkiHubHTTPError, AnkiHubRequestException
+from ...gui.operations import AddonQueryOp
+from ...settings import config
+
+
+def _fetch_username_in_background() -> None:
+    username = ""
+    if config.is_logged_in() and not config.username():
+        client = AnkiHubClient()
+        try:
+            username = client.get_user_details()["username"]
+        except (AnkiHubRequestException, AnkiHubHTTPError) as exc:
+            LOGGER.error(f"Failed to fetch username: {exc}")
+    config.save_username(username)
+
+
+def fetch_username_in_background() -> None:
+    def on_done(_: None) -> None:
+        LOGGER.info("Fetched username.")
+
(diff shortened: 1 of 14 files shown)


abdnh:
Thanks @RisingOrange!

Some other tests are failing to a missing mock address.

We're using `requests_mock.get("https://app.ankihub.net/api/users/me", json=data}` in some tests. I hope we don't need to add that to every single test that touches `fetch_username_in_background()`

ankihub/gui/ankiaddonconfig
RisingOrange, line 1: :wrench: This change should be removed

ankihub/gui/operations/username.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def _fetch_username_in_background() -> None:
+    username = ""
+    if config.is_logged_in() and not config.username():
+        client = AnkiHubClient()
+        try:
+            username = client.get_user_details()["username"]
+        except (AnkiHubRequestException, AnkiHubHTTPError) as exc:
+            LOGGER.error(f"Failed to fetch username: {exc}")
```
RisingOrange, line 17: :wrench: Let's no use log level "error" here, because then it will be sent to sentry because of Sentry's logging integration.  Fetching the user name can fail when the token is expired and we don't want issues on sentry in this case.
```suggestion
            LOGGER.warning(f"Failed to fetch username: {exc}")
```

We are using a logging level in `_setup_feature_flags`, but feature flags can be fetched without being logged in (which I find a bit weird), so for them it's really an exception when they can't be fetched.

RisingOrange (changes requested):
Looks good overall, requested some changes.

I like that we will fetch the user name now. I don't think it's really needed for filtering out note types which were already published (see comment below), but it's good that we will have more user names in the note type names now instead of emails.

RisingOrange:
This is not related to excluding note types with duplicate names, but we should make this change to `note_type_management.add_note_type`:
```diff
diff --git a/ankihub/main/note_type_management.py b/ankihub/main/note_type_management.py
index 08428eec..fe6dae6c 100644
--- a/ankihub/main/note_type_management.py
+++ b/ankihub/main/note_type_management.py
@@ -20,7 +20,11 @@ def add_note_type(ah_did: uuid.UUID, note_type: NotetypeDict) -> NotetypeDict:
     new_note_type = aqt.mw.col.models.get(NotetypeId(new_mid))
     # Send base name to AnkiHub, as it will take care of adding the deck name and username
     new_note_type["name"] = note_type["name"]
-    new_name = client.create_note_type(ah_did, new_note_type)["name"]
+    try:
+        new_name = client.create_note_type(ah_did, new_note_type)["name"]
+    except Exception as e:
+        aqt.mw.col.models.remove(new_mid)
+        raise e
     new_note_type["name"] = new_name
     aqt.mw.col.models.update_dict(new_note_type)
     new_note_type = aqt.mw.col.models.get(NotetypeId(new_mid))
```

RisingOrange (approved):

PR #1117, opened 2025-03-07
[BUILD-1092] Allow extra local note type fields

RisingOrange:
Both, the AnkiHub add-on and the AnKing note types add-on, can update the fields of note types. We need to prevent issues where one add-on adds a field, and the other removes it, leading to loss of note fields contents for that field.

We can do this by changing the add-ons so that they only add fields, but don’t remove them. 
The add-ons should just add missing fields and ignore extra fields.
This will also have the effect that users will be able to add extra fields to the note type, without the fields getting reset during sync.

## Related issues
https://ankihub.atlassian.net/jira/software/c/projects/BUILD/boards/1?selectedIssue=BUILD-1092

## Proposed changes
- When adjusting fields of note types in the AnkiHubImporter, only add fields and don’t remove them
  - Don’t require “ord” values to be the same between the local and the remote note type
  - Make sure “ord” values are adjusted properly
- Adjust code for exporting notes to ignore extra fields

### Syncing with extra fields
- Add a field to an AnkiHub note type (in Anki) and move it before the `ankihub_id` field
- Sync with AnkiHub
- Check that the field was not removed by the sync
- Now move the new field after the `ankihub_id` field
- Check that the `ankihub_id` field was moved so that it's the last field again
- Field values should be in the correct fields throughout

### Making suggestions
- Creating suggestions from notes where the note type has extra local fields should work as previously without problems
  - The extra local fields should be ignored

### DIFF
--- a/tests/addon/test_integration.py
+++ b/tests/addon/test_integration.py
@@ -1669,7 +1669,7 @@ class TestSuggestNotesInBulk:
             assert bulk_suggestions_method_mock.call_count == 1
 
             def get_ah_nids_from_suggestions(
-                suggestions: List[Union[ChangeNoteSuggestion, NewNoteSuggestion]]
+                suggestions: List[Union[ChangeNoteSuggestion, NewNoteSuggestion]],
             ) -> List[uuid.UUID]:
                 return [suggestion.ah_nid for suggestion in suggestions]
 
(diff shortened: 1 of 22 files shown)


tests/addon/test_integration.py
```diff
@@ -1693,54 +1694,161 @@ def get_ah_nids_from_suggestions(
...
+                [ANKIHUB_NOTE_TYPE_FIELD_NAME, "Extra", "Text"],
+                ["Text", ANKIHUB_NOTE_TYPE_FIELD_NAME],
+                ["Text", "Extra", ANKIHUB_NOTE_TYPE_FIELD_NAME],
+            ),
+            # Local-only field positioned before remote field (no adjustment needed)
+            (
+                ["Extra", "Text", ANKIHUB_NOTE_TYPE_FIELD_NAME],
+                ["Text", ANKIHUB_NOTE_TYPE_FIELD_NAME],
+                ["Extra", "Text", ANKIHUB_NOTE_TYPE_FIELD_NAME],
+            ),
```
abdnh, line 1761: Can we add some tests for fields differing only in case?

abdnh (approved):
LGTM

RisingOrange:
@abdnh 
> (Is the git history messed up?)

The PR was based on a different PR branch (or multiple), but it shouldn't cause any problems

abdnh (approved):

PR #1118, opened 2025-03-07
[BUILD-1079] Support importing notes with subset of fields

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/BUILD-1079
https://ankihub.atlassian.net/browse/BUILD-1078

## Proposed changes

After BUILD-1078, the web app can return notes with missing fields (indicating empty fields). This PR implements the necessary add-on changes to handle this correctly.

## How to reproduce

- Add a new field to an AnkiHub note type then publish the field from the Deck Management screen.
- Log in and sync with AnkiHub in a new Anki profile.
- Make any change to the new field in an existing note then sync again.
- Notice your changes are cleared after sync.
- Same process can be done with the Reset local changes option to confirm missing fields are cleared.

### DIFF
--- a/ankihub/db/db.py
+++ b/ankihub/db/db.py
@@ -134,14 +134,16 @@ class _AnkiHubDB:
         upserted_notes: List[NoteInfo] = []
         note_dicts = []
         for note_data in notes_data:
-
             # Prepare fields and tags for insertion
-            fields = join_fields(
-                [
-                    field.value
-                    for field in sorted(note_data.fields, key=lambda field: field.order)
-                ]
-            )
+            field_values = []
+            for (_, field_name) in self.note_type_field_ords_and_names(
+                ankihub_did=ankihub_did, anki_note_type_id=NotetypeId(note_data.mid)
+            ):
+                field = next(
+                    (f for f in note_data.fields if f.name == field_name), None
+                )
+                field_values.append(field.value if field else "")
+            fields = join_fields(field_values)
             tags = " ".join([tag for tag in note_data.tags if tag is not None])
 
             note_dicts.append(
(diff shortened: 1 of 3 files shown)


abdnh:
(Looking into the failing test)

tests/addon/test_integration.py
```diff
@@ -2690,6 +2690,39 @@ def import_note(note_info: NoteInfo) -> AnkiHubImportResult:
                 assert len(import_result.updated_nids) == 0
                 assert len(import_result.deleted_nids) == 0
 
+    def test_import_note_with_missing_fields(
```
RisingOrange, line 2693: :wrench: It would be good to have a test where the note has a value in a field and then the fields gets cleared after importing a note where the field was missing.
So the same as here, but the note already exists in Anki and has some text in the "Back" field before the import.

ankihub/db/db.py
```diff
@@ -638,9 +643,15 @@ def note_type_field_names(
...
-            )["flds"]
+            for field in sorted(
+                (
+                    field
+                    for field in self.note_type_dict(
+                        ankihub_did=ankihub_did, note_type_id=anki_note_type_id
+                    )["flds"]
+                ),
+                key=lambda f: f["ord"],
+            )
```
RisingOrange, line 654: :question: Is this change necessary? The fields should already be ordered. We were relying on this in `_AnkiHubDB.note_data`: https://github.com/AnkiHubSoftware/ankihub_addon/blob/df21cbbf71bccc6a3c6ba14dcb3c4acd1b3b666c/ankihub/db/db.py#L304

RisingOrange (commented):
Looks good!

ankihub/db/db.py
abdnh: Hmm doesn't seem to be strictly necessary but I just wanted to keep the explicit sorting done above in upsert_notes_data() and thought it would be better to add it here instead.

RisingOrange (approved):

PR #1119, opened 2025-03-09
Change primary key of AnkiHubNoteType to the Anki note type id

RisingOrange:
## Related issues

- [x] Closes #

## Proposed changes
Describe the big picture of your changes here to communicate to the reviewers why we should accept this pull request. If it fixes a bug or resolves a feature request, be sure to link to that issue.

## How to reproduce
Explain like if someone who doesn't know this project is reviewing your changes and how they can replicate this.

## Screenshots and videos

Paste here any screenshots or videos related to your changes, if applicable.

## Further comments
If this is a relatively large or complex change, kick off the discussion by explaining why you chose the solution you did and what alternatives you considered, etc...

### DIFF
--- a/ankihub/db/db.py
+++ b/ankihub/db/db.py
@@ -297,7 +297,7 @@ class _AnkiHubDB:
             return None
 
         field_names = self.note_type_field_names(
-            ankihub_did=note.ankihub_deck_id, anki_note_type_id=note.anki_note_type_id
+            anki_note_type_id=note.anki_note_type_id
         )
 
         return self._build_note_info(note, {note.anki_note_type_id: field_names})
@@ -317,7 +317,6 @@ class _AnkiHubDB:
         for note in notes:
             if note.anki_note_type_id not in field_names_by_mid:
                 field_names_by_mid[note.anki_note_type_id] = self.note_type_field_names(
-                    ankihub_did=cast(uuid.UUID, note.ankihub_deck_id),
                     anki_note_type_id=cast(NotetypeId, note.anki_note_type_id),
                 )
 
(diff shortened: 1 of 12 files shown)


ankihub/db/utils.py
```diff
@@ -6,10 +6,10 @@ class TimedLock:
     """Thread lock with a timeout. Raises a RuntimeError if the lock could not be acquired within the timeout."""
 
     def __init__(self, timeout_seconds: float):
-        self.lock: threading.Lock = threading.Lock()
+        self.lock: threading.RLock = threading.RLock()
```
RisingOrange, line 9: Changed this into a re-entrant lock because `_AnkiHubDB.remove_deck()` is now calling _AnkiHubDB.remove_note_types_of_deck()`, and both of them use `_AnkiHubDB.write_lock`. Without this change there would be a deadlock in `_AnkiHubDB.remove_deck()`

abdnh:
> Some users might have multiple note types with the same note type id but different deck ids in the database, which would lead to primary key conflicts.

I was curious about testing this case, but found no note types with the same Anki IDs on AnkiHub.

RisingOrange:
> > Some users might have multiple note types with the same note type id but different deck ids in the database, which would lead to primary key conflicts.
> 
> I was curious about testing this case, but found no note types with the same Anki IDs on AnkiHub.

Yes, the `NoteType.anki_id` field has now a unique constraint on the web app:
https://github.com/ankipalace/ankihub/blob/1f591ec144bf7fc2b63fff8aa8d5399fe0e305ab/ankihub/decks/models.py#L67

The problem is that some note types are used by notes in multiple decks. (This will be fixed by https://github.com/AnkiHubSoftware/ankihub/pull/2589)
Because of this, the same note type was inserted into the AnkiHub DB multiple times for some users, with the same note type id, but different ankihub deck ids

PR #1121, opened 2025-03-11
fix: Publish note type options

RisingOrange:
Fixes issues found during QA: https://ankihub.atlassian.net/browse/NRT-13?focusedCommentId=16567

### DIFF
--- a/ankihub/db/db.py
+++ b/ankihub/db/db.py
@@ -613,6 +613,13 @@ class _AnkiHubDB:
             .scalar()
         )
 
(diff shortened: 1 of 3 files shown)


abdnh (approved):

RisingOrange:
@abdnh We are still not excluding some note types which could be excluded, because of note type names only differing by e.g. `-7cb94`. But it I think this should be good enough

abdnh (approved):
Nice catch

PR #1122, opened 2025-03-11
[NRT-48] fix: Refresh Anki UI after ankihub sync

RisingOrange:
There is a bug that can lead to field values of a note ending up in the wrong field after an AnkiHub sync.

Also, after syncing with AnkiHub, changes are not displayed in the deck browser until you click on a deck or on the "Decks" button.

## Related issues
https://ankihub.atlassian.net/jira/software/c/projects/NRT/boards/41?selectedIssue=NRT-48

## Proposed changes
- Add `refresh_anki_ui()` function to `gui.utils`
- Refresh UI after the sync
- Use `refresh_anki_ui()` in other places where we need to refresh Anki's UI

## How to reproduce the bug
- Change field order of AnkiHub note type
- Open a field that has the note type in the editor
- Sync with AnkiHub
- Click a field of the note in the editor
- Select some other note
- Select the original note
- Field values are mixed up!

This happens because the field order in the editor doesn't get updated after the AnkiHub sync. When you click on a field of the note in the editor, you trigger the editor to save the note - but it saves the fields incorrectly, because they are in the wrong order.


ankihub/gui/operations/ankihub_sync.py
```diff
@@ -223,6 +223,8 @@ def _on_sync_done(on_done: Callable[[Future], None]) -> None:
     LOGGER.info("Sync with AnkiHub done.")
     config.log_private_config()
 
+    QTimer.singleShot(500, lambda: refresh_anki_ui())
```
RisingOrange, line 226: @abdnh 

I added the delay here in order to prevent the deck browser from fading out:
https://github.com/user-attachments/assets/7ba2fdb6-fa37-4a2f-b19b-f686b62fefcc

With the delay I don't get a fade-out:
https://github.com/user-attachments/assets/2d7f4ce9-3014-4e10-a6a2-30b212d4ee8e

The fade-out happens because the main window is not focused when `refresh_anki_ui()` is called. (I think it's because the progress dialog of the sync didn't close yet)
https://github.com/ankitects/anki/blob/9b5da546be49f37c8d6c286e09c86074b2f0c278/qt/aqt/main.py#L869-L870

Any ideas on how to improve this?

RisingOrange: I think I found a good solution using the `focus_did_change` hook

ankihub/gui/utils.py
```diff
@@ -774,3 +775,37 @@ def get_ah_did_of_deck_or_ancestor_deck(anki_did: DeckId) -> Optional[uuid.UUID]
 
 def logged_into_ankiweb() -> bool:
     return bool(aqt.mw.pm.sync_auth())
+
+
+def refresh_anki_ui() -> None:
+    changes = OpChanges(
```
abdnh, line 781: I would simply call `mw.reset()` here to refresh everything so we don't have to worry about figuring which fields to set and hooks to call. Anki also calls `mw.update_undo_actions()`, which I think is needed because we clear the undo queue when we do direct DB write operations in some tasks.

ankihub/gui/utils.py
```diff
@@ -774,3 +775,37 @@ def get_ah_did_of_deck_or_ancestor_deck(anki_did: DeckId) -> Optional[uuid.UUID]
...
+        notetype=True,
+        note_text=True,
+        browser_table=True,
+        browser_sidebar=True,
+    )
+    # Anki's UI componennts subscribe to this hook to refresh themselves when needed.
+    # This is the mechanism that Anki uses to update the UI when its backend makes changes to the collection.
+    gui_hooks.operation_did_execute(
```
abdnh, line 789: Anki also calls `gui_hooks.state_did_reset()` here

ankihub/gui/utils.py
```diff
@@ -774,3 +775,37 @@ def get_ah_did_of_deck_or_ancestor_deck(anki_did: DeckId) -> Optional[uuid.UUID]
...
+        changes=changes,
+        handler=None,
+    )
+
+    # Study queue related UI is refreshed when focus is not None.
+    # This prevents the deck browser from fading out when the main window temporarily loses focus.
```
abdnh, line 795: I remember seeing the fade-out effect getting stuck sometimes but I can't reproduce it now.

Does calling `mw.fade_in_webview()` instead of constructing an OpChanges here work? That's more intuitive and less reliant on Anki's internal handling.

ankihub/gui/utils.py
RisingOrange: Hm, that might be better. However, when I replace `refresh_anki_ui()` with `aqt.mw.reset()`, Anki consistently lags:

https://github.com/user-attachments/assets/e6d87d6d-230f-4dcd-962e-16ea460b44ae

It doesn't happen with `refresh_anki_ui()`:

https://github.com/user-attachments/assets/a9122a35-b7cf-4b46-b7f9-ab83e38bf724

RisingOrange: It seems like the `update_undo_actions()` is causing the lag. I'll try to find out why

ankihub/gui/utils.py
RisingOrange: > I remember seeing the fade-out effect getting stuck sometimes but I can't reproduce it now.#

You mean on this PR branch?

ankihub/gui/utils.py
abdnh: Probably due to the backend call in _check_backend_undo_status()?

ankihub/gui/utils.py
abdnh: I mean generally in Anki when progress dialogs are shown. Didn't manage to reproduce it in this branch yet.

ankihub/gui/utils.py
RisingOrange: Yes, _`backed.get_undo_status()` takes a long time:
image

Any idea why or how we can improve this?

RisingOrange: Oh, I think it might be because we are accessing the Anki DB to get review summaries/data at the same time

abdnh: Maybe moving the .run_in_background() calls after the refresh_anki_ui() call will help?

RisingOrange: Hm, but the lag still happens when the browser is open while you sync

RisingOrange:
@abdnh Client tests will pass once https://github.com/AnkiHubSoftware/ankihub/pull/2619 is merged. Can you review it please?

abdnh (approved):

PR #1131, opened 2025-03-24
[NRT-98] Fix suggestion fields being mixed up

abdnh:
`_fields_that_changed()` compares different fields in some cases because `to_note_data()` doesn't include empty fields.

## Related issues

https://ankihub.atlassian.net/browse/NRT-98

## Proposed changes

- Fix `_fields_that_changed()` so that it matches fields by their names.
- Add the `include_empty_fields` parameter to `to_note_data()` and use it in `_change_note_suggestion()`

## How to reproduce

The issue can be reproduced by making changes to a field that has empty fields before it. Because empty fields were removed from the fields list before comparison and `_fields_that_changed()` doesn't expect that, it was resulting in the wrong fields being reported as changed.

### DIFF
--- a/ankihub/main/suggestions.py
+++ b/ankihub/main/suggestions.py
@@ -338,7 +338,7 @@ def _new_note_suggestion(
 def _change_note_suggestion(
     note: Note, change_type: SuggestionType, comment: str
 ) -> Optional[ChangeNoteSuggestion]:
-    note_from_anki_db = to_note_data(note)
+    note_from_anki_db = to_note_data(note, include_empty_fields=True)
     assert isinstance(note_from_anki_db, NoteInfo)
     assert note_from_anki_db.ah_nid is not None
     assert note_from_anki_db.tags is not None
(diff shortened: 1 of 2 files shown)


ankihub/main/suggestions.py
```diff
@@ -383,11 +383,13 @@ def _added_and_removed_tags(
...
-        for cur_field, prev_field in zip(cur_fields, prev_fields)
-        if cur_field.value != prev_field.value
-    ]
+    result = []
+    for prev_field in prev_fields:
+        cur_field = next(
+            (field for field in cur_fields if field.name == prev_field.name), None
+        )
+        if cur_field and cur_field.value != prev_field.value:
+            result.append(cur_field)
```
RisingOrange, line 392: :pick: I think it would be better to undo this change. It's not needed now that both field lists contain all fields, including empty ones.

abdnh: We allow extra local fields though? What if the deck owner has extra fields they didn't publish yet?

RisingOrange: `to_note_data` only returns fields which are defined in the note type. 

The fields returned by `to_note_data` come from here:
https://github.com/AnkiHubSoftware/ankihub_addon/blob/6b1bfe24b9bebadafafdd66d87967bdec502cdec/ankihub/main/exporting.py#L48-L58

RisingOrange (approved):

PR #1134, opened 2025-04-01
[ACTV-28] Fix blank pages and crashes when switching sidebar buttons

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/ACTV-28

## Proposed changes

The add-on's use of a separate QWebEnginePage for each sidebar button was causing pages to fail to load and crashes after doing a few reviews.

## How to reproduce

Steps to reproduce the issues without this PR:
1. Open the chatbot.
2. Do some reviews.
3. Switch to other sidebar buttons a few times. You should see the FirstAidForward page specifically sometimes fails to load. When that happens, and you switch to a different button, Anki will crash.

## Further comments

The crash happens inside the `AnkiWebView.setPage()` call in `ReviewerSidebar.set_content_url()`. This is probably a Qt bug.

### DIFF
--- a/ankihub/gui/reviewer.py
+++ b/ankihub/gui/reviewer.py
@@ -144,6 +144,8 @@ class ReviewerSidebar:
         page.setBackgroundColor(theme_manager.qcolor(colors.CANVAS))
         aqt.qconnect(page.loadFinished, self._on_content_page_loaded)
         self.content_webview.setPage(page)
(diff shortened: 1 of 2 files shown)


augdiebold (approved):

abdnh:
@augdiebold can you merge this? I can't merge it due to a permission issue

PR #1145, opened 2025-05-13
Fix client tests on CI

RisingOrange:
This PR fixes client tests on CI by waiting for the server to become responsive before running a test.


ddevdan (approved):
LGTM

PR #1172, opened 2025-07-18
[ACTV-128] Add more logging to debug "AnkiHub Sync Cancelled" issue

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/ACTV-128

## Proposed changes

This adds logging to ask_user_dialog() calls done by Anki as part of full syncs to help us debug the "AnkiHub Sync Cancelled" issue.

## How to reproduce

Triggers a full sync from the option under Preferences > Syncing and check the confirmation dialog appears normally and that AnkiHub is logging the call.

### DIFF
--- a/ankihub/gui/operations/ankihub_sync.py
+++ b/ankihub/gui/operations/ankihub_sync.py
@@ -2,7 +2,7 @@ from concurrent.futures import Future
 from dataclasses import dataclass
 from datetime import timedelta
 from functools import partial
-from typing import Callable, List, Optional
+from typing import Any, Callable, List, Optional
 from uuid import UUID
 
 import aqt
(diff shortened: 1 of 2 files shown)


abdnh:
> This changes the calling convention, causing "multiple values for argument" errors, especially when functions also use VAR_POSITIONAL parameters.

Having POSITIONAL_OR_KEYWORD args added to new_args will trigger the "multiple values for argument" error in ask_user_dialog() because `callback` is removed and a different POSITIONAL_OR_KEYWORD arg is interpreted as `callback`. Appending to `new_kwargs` instead works here and I couldn't reproduce the same issue with it.

r-lelis (approved):

PR #1176, opened 2025-07-23
[INTP-141] Avoid comitting vendored dependencies

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/INTP-141

## Proposed changes

This removes vendored packages committed to the `lib` directory and use pyproject.toml and uv to install them at build time instead.

## How to reproduce

- Run scripts/release.sh
- Install the produced ankihub.ankiaddon and confirm add-on is running normally.


abdnh:
> The build script fails because it attempts to install dependencies using `uv pip install --group bundle`. The `--group` flag is not supported by `uv pip install` (it is intended for `uv sync` or `uv install`).

It's supported in recent uv releases: https://github.com/astral-sh/uv/commit/ba732311641f6eeda8a2f09fb83ac1b81a5e17db

RisingOrange (commented):
❓ @abdnh Can you explain the approach a bit more? I see that some packages were removed from the `lib` directory, but not all.

It's nice that we are using a fork of `mashumaro` now!

abdnh:
>  I see that some packages were removed from the lib directory, but not all.

That's just a mistake.

.github/workflows/create_release.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - name: Install deps
         run: |
-          uv sync --locked --dev --group production
+          uv sync --locked --dev --group production --group bundle
```
RisingOrange, line 51: ⛏️ I think this change isn't needed, as well as the changes to this command in other files.

The release script is called right after:
https://github.com/AnkiHubSoftware/ankihub_addon/blob/5b6e0c16aa70ce64c2451c5f63aebf3cb1bffb3d/.github/workflows/create_release.yml#L55

This script runs the `build.py` script, and the `build.py` script installs the dependencies from the bundle group into the `lib` directory.

By the way, we have an `uv sync` command in the `action.yml` file, which doesn't include the bundle group.
https://github.com/AnkiHubSoftware/ankihub_addon/blob/5b6e0c16aa70ce64c2451c5f63aebf3cb1bffb3d/.github/actions/setup-addon/action.yml#L21

This is used in the test setup, and the tests still work because the `build.py` script is called after the `uv sync` command.

It's a bit weird to install the dependencies from the `bundle` group both to the python env **and** to the `lib` folder. I think the easiest solution would be to just install the dependencies to the `lib` directory, and not to the env. 
Alternatively, we could use the dependencies from the env during development, and just install the dependencies into the `dist/release/lib` folder when creating the `.ankiaddon` file.

RisingOrange (approved):
We can remove the code for adding the `lib` folder to the system path: https://github.com/AnkiHubSoftware/ankihub_addon/blob/5b6e0c16aa70ce64c2451c5f63aebf3cb1bffb3d/ankihub/__init__.py#L5-L10

RisingOrange (approved):

PR #1266, opened 2026-04-02
[ACTV-289] Trigger onboarding tour on first signup

cjasm:
## Related issues

- [x] Closes # [ACTV-289](https://ankihub.atlassian.net/browse/ACTV-289)

## Proposed changes
Triggers the Onboarding Tour when (new) users sign in or when they open up the addon without having syncing.

## How to reproduce

Test Case 1:
1. Create a new account
2. "Sign In" in the AnkiHub addon
3. The onboarding tour should appear
4. Click on "Maybe later"
5. Close the Anki and open it again
6. The onboarding tour should appear

Test Case 2:
1. Create a new account
2. "Sign In" in the AnkiHub addon
3. The onboarding tour should appear
4. Click on "X"
5. Close the Anki and open it again
6. The onboarding tour should not appear

[ACTV-289]: https://ankihub.atlassian.net/browse/ACTV-289?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/operations/ankihub_sync.py
+++ b/ankihub/gui/operations/ankihub_sync.py
@@ -248,6 +248,15 @@ def _schedule_post_sync_tasks() -> None:
     aqt.mw.taskman.run_in_background(send_review_data, on_done=_on_send_review_data_done)
     _maybe_send_daily_review_summaries()
     aqt.mw.taskman.run_on_main(maybe_show_subdeck_due_date_reminders)
+    aqt.mw.taskman.run_on_main(_show_onboarding_prompt_if_first_sync)
+
+
+def _show_onboarding_prompt_if_first_sync() -> None:
+    if config.last_deck_sync() is None:
+        from ..tutorial import prompt_for_onboarding_tutorial
+
+        prompt_for_onboarding_tutorial()
+        config.update_last_deck_sync()
 
 
 def _on_clear_unused_tags_done(future: Future) -> None:
--- a/ankihub/gui/operations/ankihub_sync.py
+++ b/ankihub/gui/operations/ankihub_sync.py
@@ -248,15 +248,6 @@ def _schedule_post_sync_tasks() -> None:
     aqt.mw.taskman.run_in_background(send_review_data, on_done=_on_send_review_data_done)
     _maybe_send_daily_review_summaries()
     aqt.mw.taskman.run_on_main(maybe_show_subdeck_due_date_reminders)
-    aqt.mw.taskman.run_on_main(_show_onboarding_prompt_if_first_sync)
-
-
-def _show_onboarding_prompt_if_first_sync() -> None:
-    if config.last_deck_sync() is None:
-        from ..tutorial import prompt_for_onboarding_tutorial
-
-        prompt_for_onboarding_tutorial()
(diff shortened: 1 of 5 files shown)


ankihub/gui/operations/ankihub_sync.py
```diff
@@ -248,15 +248,6 @@ def _schedule_post_sync_tasks() -> None:
     aqt.mw.taskman.run_in_background(send_review_data, on_done=_on_send_review_data_done)
     _maybe_send_daily_review_summaries()
     aqt.mw.taskman.run_on_main(maybe_show_subdeck_due_date_reminders)
-    aqt.mw.taskman.run_on_main(_show_onboarding_prompt_if_first_sync)
-
-
-def _show_onboarding_prompt_if_first_sync() -> None:
```
abdnh, line 254: I think we should keep this because users can also log in to AnkiHub using _Tools > Preferences > Third-Party Services_ and the tour won't be triggered there.

abdnh (changes requested):

ankihub/gui/operations/ankihub_sync.py
cjasm: Could you take a look on the changes?

I reverted this changed but trigger the onboarding only when not triggered on signin

ankihub/settings.py
```diff
@@ -219,6 +219,7 @@ class PrivateConfig(DataClassJSONMixin):
     user_details: dict = field(default_factory=dict)
     block_exams_subdecks: List[BlockExamSubdeckConfig] = field(default_factory=list)
     onboarding_tutorial_pending: bool = False
+    onboarding_tutorial_show_on_sync: bool = True
```
abdnh, line 222: Do we need an additional bool? Can't we simply use update_last_deck_sync() in _maybe_show_onboarding_tutorial_after_login() for simpler state management?

cjasm: Yes, this boolean avoids displaying the tour twice "in a row".

It controls whether the tour is shown on login, then does not display the same tour on sync. In case, the user trigger the login through "thirdy-party", it displays the tour when sync.

abdnh: My point is that adding an update_last_deck_sync() call after the login trigger should prevent that. I can't see how the tour can be triggered twice in that case. Or am I missing something?

cjasm: The user can click on "maybe later" and not want to sync at the moment the tour was shown. 

According to the design, the tour should be displayed again once they open Anki, which could not occur if I just control through the last deck sync.

abdnh: > The user can click on "maybe later" and not want to sync at the moment the tour was shown.

OK, that makes sense if the tour prompt is supposed to show again when they sync without restarting Anki.

> According to the design, the tour should be displayed again once they open Anki, which could not occur if I just control through the last deck sync.

That's already handled by `onboarding_tutorial_pending`.

abdnh (dismissed):

abdnh (approved):

PR #1278, opened 2026-04-09
[ACTV-290][Bugfix] Onboarding tour locks on decks with all cards suspended

cjasm:
## Related issues

- [x] Closes # [ACTV-290](https://ankihub.atlassian.net/browse/ACTV-290)

## Proposed changes
Unsuspend the cards when all cards of the getting started deck are suspended

#### Setup
1. Must have the Getting started deck
2. Must suspend all cards from the deck

#### Test Case
1. Trigger the onboarding tour in the help menu
2. The tutorial is triggered
3. The second step should display the new message and unsuspend the cards
4. The rest of the tour should work properly

[ACTV-290]: https://ankihub.atlassian.net/browse/ACTV-290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/tutorial.py
+++ b/ankihub/gui/tutorial.py
@@ -3,7 +3,7 @@ import json
 from asyncio.futures import Future
 from dataclasses import dataclass
 from functools import cached_property, partial
-from typing import Any, Callable, Dict, List, Optional, Sequence, Set, Tuple, TypedDict, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Set, Tuple, TypedDict, Union, cast
 
 import aqt
 from anki.cards import CardId
@@ -41,9 +41,11 @@ from typing_extensions import Required, Unpack
 
 from .. import LOGGER
 from ..addon_ankihub_client import AddonAnkiHubClient as AnkiHubClient
+from ..db import ankihub_db
 from ..django import render_template, render_template_from_string
 from ..gui.overlay_dialog import OverlayDialog, OverlayTarget
 from ..main.deck_unsubscribtion import uninstall_deck
+from ..main.reset_local_changes import reset_local_changes_to_notes
 from ..settings import config
 from .flashcard_selector_dialog import (
     show_flashcard_selector,
@@ -885,11 +887,17 @@ class OnboardingTutorial(DeckBrowserOverviewBackdropMixin, Tutorial):
         aqt.mw.moveToState("deckBrowser")
 
     def _unsuspend_cards_and_move_to_intro_deck_overview(self, on_done: Callable[[], None]) -> None:
-        intro_deck_config = config.deck_config(config.intro_deck_id)
-        cids: Sequence[CardId] = aqt.mw.col.find_cards(f'deck:"{intro_deck_config.name}"')
+        ah_did = config.intro_deck_id
+        intro_deck_config = config.deck_config(ah_did)
+        cids = list(aqt.mw.col.decks.cids(intro_deck_config.anki_id, True))
 
         if cids:
-            aqt.mw.col.sched.unsuspend_cards(ids=list(cids))
+            aqt.mw.col.sched.unsuspend_cards(ids=cids)
+            aqt.mw.col.sched.schedule_cards_as_new(card_ids=cids, restore_position=True)
+
+        if not self._has_cards_to_review():
+            nids = ankihub_db.anki_nids_for_ankihub_deck(ah_did)
+            reset_local_changes_to_notes(nids=nids, ah_did=ah_did)
 
         self._move_to_intro_deck_overview(on_done)
 
--- a/ankihub/gui/tutorial.py
+++ b/ankihub/gui/tutorial.py
@@ -10,6 +10,7 @@ from anki.cards import CardId
 from anki.config import Config
 from anki.hooks import wrap
 from anki.notes import NoteId
+from anki.scheduler.v3 import Scheduler
 from aqt import gui_hooks
 from aqt.browser import Browser
 from aqt.browser.sidebar.item import SidebarItem, SidebarItemType
@@ -893,6 +894,7 @@ class OnboardingTutorial(DeckBrowserOverviewBackdropMixin, Tutorial):
         self._move_to_intro_deck_overview(on_done)
 
     def _has_cards_to_review(self) -> bool:
+        assert isinstance(aqt.mw.col.sched, Scheduler)
         return bool(aqt.mw.col.sched.get_queued_cards().cards)
 
     @cached_property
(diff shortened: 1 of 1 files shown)


ankihub/gui/tutorial.py
```diff
@@ -882,6 +883,19 @@ def _move_to_deck_browser(self, on_done: Callable[[], None]) -> None:
...
+        cids: Sequence[CardId] = aqt.mw.col.find_cards(f'deck:"{intro_deck_config.name}"')
+
+        if cids:
+            aqt.mw.col.sched.unsuspend_cards(ids=list(cids))
+
+        self._move_to_intro_deck_overview(on_done)
+
+    def _has_cards_to_review(self) -> bool:
+        assert isinstance(aqt.mw.col.sched, Scheduler)
+        return bool(aqt.mw.col.sched.get_queued_cards().cards)
```
abdnh, line 923: Pushed a fix for the mypy issue. Will review this tomorrow

ankihub/gui/tutorial.py
```diff
@@ -882,6 +883,19 @@ def _move_to_deck_browser(self, on_done: Callable[[], None]) -> None:
         self._monitor_mw_state_change(on_done)
         aqt.mw.moveToState("deckBrowser")
 
+    def _unsuspend_cards_and_move_to_intro_deck_overview(self, on_done: Callable[[], None]) -> None:
+        intro_deck_config = config.deck_config(config.intro_deck_id)
+        cids: Sequence[CardId] = aqt.mw.col.find_cards(f'deck:"{intro_deck_config.name}"')
+
+        if cids:
+            aqt.mw.col.sched.unsuspend_cards(ids=list(cids))
```
abdnh: As noted in QA, unsuspending the cards is not enough. We also need to reset them as new cards:
```diff
diff --git a/ankihub/gui/tutorial.py b/ankihub/gui/tutorial.py
index 7a996207..7b13bc71 100644
--- a/ankihub/gui/tutorial.py
+++ b/ankihub/gui/tutorial.py
@@ -883,13 +883,12 @@ class OnboardingTutorial(DeckBrowserOverviewBackdropMixin, Tutorial):
         self._monitor_mw_state_change(on_done)
         aqt.mw.moveToState("deckBrowser")

-    def _unsuspend_cards_and_move_to_intro_deck_overview(self, on_done: Callable[[], None]) -> None:
+    def _reset_cards_and_move_to_intro_deck_overview(self, on_done: Callable[[], None]) -> None:
         intro_deck_config = config.deck_config(config.intro_deck_id)
-        cids: Sequence[CardId] = aqt.mw.col.find_cards(f'deck:"{intro_deck_config.name}"')
-
+        cids = list(aqt.mw.col.find_cards(f'deck:"{intro_deck_config.name}"'))
         if cids:
-            aqt.mw.col.sched.unsuspend_cards(ids=list(cids))
-
+            aqt.mw.col.sched.unsuspend_cards(ids=cids)
+            aqt.mw.col.sched.schedule_cards_as_new(card_ids=cids, restore_position=True)
         self._move_to_intro_deck_overview(on_done)

     def _has_cards_to_review(self) -> bool:
@@ -919,7 +918,7 @@ class OnboardingTutorial(DeckBrowserOverviewBackdropMixin, Tutorial):
                     "This deck was already reviewed. Click on <b>Next</b> and we'll "
                     "bring the cards back so you can continue with the tour."
                 )
-                next_callback = self._unsuspend_cards_and_move_to_intro_deck_overview
+                next_callback = self._reset_cards_and_move_to_intro_deck_overview

             steps.append(
                 TutorialStep(
```

ankihub/gui/tutorial.py
```diff
@@ -882,6 +883,19 @@ def _move_to_deck_browser(self, on_done: Callable[[], None]) -> None:
         self._monitor_mw_state_change(on_done)
         aqt.mw.moveToState("deckBrowser")
 
+    def _unsuspend_cards_and_move_to_intro_deck_overview(self, on_done: Callable[[], None]) -> None:
+        intro_deck_config = config.deck_config(config.intro_deck_id)
+        cids: Sequence[CardId] = aqt.mw.col.find_cards(f'deck:"{intro_deck_config.name}"')
```
abdnh:
```suggestion
        cids = aqt.mw.col.decks.cids(intro_deck_config.anki_id, True)
```

Using the name can be unreliable (because the actual name can be different as we saw while testing yesterday).

ankihub/gui/tutorial.py
```diff
@@ -882,6 +883,19 @@ def _move_to_deck_browser(self, on_done: Callable[[], None]) -> None:
...
+
+        if cids:
+            aqt.mw.col.sched.unsuspend_cards(ids=list(cids))
+
+        self._move_to_intro_deck_overview(on_done)
```
abdnh, line 907: There can still be edge cases where no cards will be available (if the cards are deleted or daily limits exhausted). We should add a check here before the `_move_to_intro_deck_overview()` call:
```python
        if not self._has_cards_to_review():
            showWarning(
                "No cards found. "
                "Please reset the deck using the <i>AnkiHub > Reset all local changes to a deck</i>"
                " menu item in the browser and check your daily limits."
            )
            self.end()
            return
```

(Add an import for `aqt.utils.showWarning`)

abdnh (changes requested):

ankihub/gui/tutorial.py
cjasm: Added the `schedule_cards_as_new` and renamed the method

ankihub/gui/tutorial.py
cjasm: Done

ankihub/gui/tutorial.py
cjasm: Agreed, I followed a different direction.

Instead of asking the user to do any action, considering that this is the onboarding tour, the app does the action for them.

Thus, the app resets the local changes or bumps its daily limit.


Do you think that it solves, or it can have any other scenario that can fail where we should display a message?

abdnh: That should work. Can we move the _bump_intro_deck_daily_limits() call before reset_local_changes_to_notes() so the latter is skipped if bumping the limits is enough? (Resetting takes longer so would be nice to skip it if not needed).

cjasm: done on [aa64fc3](https://github.com/AnkiHubSoftware/ankihub_addon/pull/1278/commits/aa64fc38f62aa2e672b5671c5c438d41fdb11f27)

abdnh (dismissed):

augdiebold (approved):

PR #1286, opened 2026-05-06
[NRT-645] Emit dialog result to web; suppress 0-added confusion

ddevdan:
## Related issues
- [x] Closes [NRT-645](https://ankihub.atlassian.net/browse/NRT-645)

## Proposed changes
Companion PR to [AnkiHubSoftware/ankihub#3646](https://github.com/AnkiHubSoftware/ankihub/pull/3646).

The web side needs to know when the **Add to block exam subdeck** dialog finishes (and whether it actually added notes) so it can disable the button for the current selection. Today the dialog closes without telling the web view anything, so the web has no way to react.

This PR adds:

- **\`BlockExamSubdeckDialog.add_succeeded\`** — flag set across the four success paths (\`_on_create_subdeck\`, \`_on_add_notes\`, \`_handle_conflict_create_new\`, \`_handle_conflict_merge\`) before \`self.accept()\`. Lets the caller distinguish a real add from a cancel/reject.
- **\`js_message_handling._on_js_message\`** now connects \`dialog.finished\` to a closure that posts \`{\"blockExamSubdeckResult\": {\"status\": \"added\" | \"cancelled\"}}\` via the existing \`_post_message_to_ankihub_js\` helper. Guarded with \`sip.isdeleted(web)\` in case the user navigates away while the dialog is open.
- **Zero-added UX cleanup** — when \`add_notes_to_block_exam_subdeck\` returns 0 (all selected notes already in the chosen subdeck), the tooltip now reads \`\"All selected notes are already in '<name>'\"\` instead of the misleading \`\"0 note(s) added to '<name>'\"\`.
- **Tooltip parent fix** — \`tooltip(...)\` now passes \`parent=aqt.mw\`. Previously \`aqt.utils.tooltip\` parented to \`activeWindow()\` (the dialog), and \`self.accept()\` immediately destroyed the tooltip child.
- **Tests** in \`TestBlockExamSubdeckDialog\` cover \`add_succeeded\` toggling on success, the unchanged-on-reject path, and the zero-added tooltip text. \`test_add_to_block_exam_subdeck_pycmd\` is extended to assert the \`blockExamSubdeckResult\` payload coming back through \`_post_message_to_ankihub_js\`.

requestId routing (for addon→web result correlation) was considered and dropped — the dialog is modal so only one add can be in flight at a time, and the web-side \`pendingAdd\` gate is sufficient.

## How to reproduce
1. Run the local stack with the AnkiHub addon and the companion ankihub PR deployed.
2. Open Anki, log in, run a Smart Search on a flashcard deck.
3. Select notes, click **Add to block exam subdeck**, complete the dialog.
4. Verify the success tooltip appears in the Anki main window (not lost to the closing dialog).
5. Verify the web button transitions to **disabled**.
6. Cancel the dialog on a fresh click — verify button stays **enabled**.
7. With notes already in subdeck \"X\", reopen the dialog and merge the same notes into \"X\" — verify the tooltip reads \`\"All selected notes are already in 'X'\"\`.

Run the addon tests:
\`\`\`
cd /home/daniel/lab/ankihub_addon
just test \"-k 'block_exam or BlockExamSubdeck'\"
\`\`\`

## Screenshots and videos
_To be added._

## Further comments
- The \`tooltip(parent=aqt.mw)\` fix predates NRT-645 (the original code had the same orphan-on-close issue) but became visible because the web-side \`pendingAdd\` flow makes the tooltip behavior more important to verify.
- \`BlockExamSubdeckDialog\` stays decoupled from the web — the \`finished\` connection is wired by the caller in \`js_message_handling\`, not inside the dialog itself.

[NRT-645]: https://ankihub.atlassian.net/browse/NRT-645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/block_exam_dialog.py
+++ b/ankihub/gui/block_exam_dialog.py
@@ -468,13 +468,7 @@ class BlockExamSubdeckDialog(QDialog):
             action_source=self.action_source,
         )
 
-        self.add_succeeded = True
-        if added_count == 0:
-            tooltip(f"All selected notes are already in '{actual_name}'", parent=aqt.mw)
-        else:
-            tooltip(f"{added_count} note(s) added to '{actual_name}'", parent=aqt.mw)
-        self.accept()
-
+        self._finish_with_added(added_count, actual_name)
         aqt.mw.moveToState("deckBrowser")
 
     def _on_add_notes(self):
@@ -504,12 +498,7 @@ class BlockExamSubdeckDialog(QDialog):
             action_source=self.action_source,
         )
 
-        self.add_succeeded = True
-        if added_count == 0:
-            tooltip(f"All selected notes are already in '{self.selected_subdeck_name}'", parent=aqt.mw)
-        else:
-            tooltip(f"{added_count} note(s) added to '{self.selected_subdeck_name}'", parent=aqt.mw)
-        self.accept()
+        self._finish_with_added(added_count, self.selected_subdeck_name)
 
     def _rename_subdeck(self, old_subdeck_path: str, new_subdeck_path: str):
         """Rename an existing subdeck."""
@@ -551,13 +540,7 @@ class BlockExamSubdeckDialog(QDialog):
             action_source=self.action_source,
         )
 
-        self.add_succeeded = True
-        if added_count == 0:
-            tooltip(f"All selected notes are already in '{actual_name}'", parent=aqt.mw)
-        else:
-            tooltip(f"{added_count} note(s) added to '{actual_name}'", parent=aqt.mw)
-        self.accept()
-
+        self._finish_with_added(added_count, actual_name)
         aqt.mw.moveToState("deckBrowser")
 
     def _handle_conflict_merge(self, conflicting_name: str):
@@ -583,11 +566,14 @@ class BlockExamSubdeckDialog(QDialog):
             action_source=self.action_source,
         )
 
+        self._finish_with_added(added_count, conflicting_name)
+
+    def _finish_with_added(self, added_count: int, subdeck_name: str) -> None:
         self.add_succeeded = True
         if added_count == 0:
-            tooltip(f"All selected notes are already in '{conflicting_name}'")
+            tooltip(f"All selected notes are already in '{subdeck_name}'", parent=aqt.mw)
         else:
-            tooltip(f"{added_count} note(s) added to '{conflicting_name}'")
+            tooltip(f"{added_count} note(s) added to '{subdeck_name}'", parent=aqt.mw)
         self.accept()
 
     def _clear_layout(self):
--- a/ankihub/gui/block_exam_dialog.py
+++ b/ankihub/gui/block_exam_dialog.py
@@ -47,6 +47,7 @@ class BlockExamSubdeckDialog(QDialog):
         self.selected_subdeck_id: Optional[DeckId] = None
 
         self.action_source = ActionSource.SMART_SEARCH
+        self.add_succeeded: bool = False
 
         self.setModal(True)
         self.setMinimumWidth(440)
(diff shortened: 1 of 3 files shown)


ankihub/gui/block_exam_dialog.py
```diff
@@ -467,8 +468,11 @@ def _on_create_subdeck(self):
...
         )
 
-        # Show success message
-        tooltip(f"{added_count} note(s) added to '{actual_name}'")
+        self.add_succeeded = True
+        if added_count == 0:
+            tooltip(f"All selected notes are already in '{actual_name}'", parent=aqt.mw)
+        else:
+            tooltip(f"{added_count} note(s) added to '{actual_name}'", parent=aqt.mw)
         self.accept()
```
RisingOrange: 🔧 This 5-line block (`add_succeeded = True` + zero/non-zero tooltip branches + `self.accept()`) is now repeated four times across `_on_create_subdeck`, `_on_add_notes`, `_handle_conflict_create_new`, and `_handle_conflict_merge` — and it has already drifted: the merge copy at lines 588 and 590 is missing `parent=aqt.mw`. Without that kwarg, `aqt.utils.tooltip` parents to `activeWindow()` (the dialog), and the immediately-following `self.accept()` destroys the tooltip child — the exact failure mode the PR description calls out as fixed across all four paths.

The right fix repairs the bug *and* removes the duplication that caused it: extract a helper so each call site collapses to one line and the `parent=` is impossible to forget.

```python
def _finish_with_added(self, added_count: int, subdeck_name: str) -> None:
    self.add_succeeded = True
    if added_count == 0:
        tooltip(f"All selected notes are already in '{subdeck_name}'", parent=aqt.mw)
    else:
        tooltip(f"{added_count} note(s) added to '{subdeck_name}'", parent=aqt.mw)
    self.accept()
```

Bonus: the two `aqt.mw.moveToState("deckBrowser")` calls (present in two paths, absent in two) become more visibly intentional rather than buried in near-duplicate blocks.

ankihub/gui/js_message_handling.py
```diff
@@ -149,10 +151,31 @@ def _on_js_message(handled: Tuple[bool, Any], message: str, context: Any) -> Any
...
         elif search_string.strip():
-            note_ids = list(aqt.mw.col.find_notes(search_string))
+            try:
+                note_ids = list(aqt.mw.col.find_notes(search_string))
+            except Exception:
+                LOGGER.warning(
+                    "ADD_TO_BLOCK_EXAM_SUBDECK: invalid search string, treating as empty",
+                    search_string=search_string,
+                )
+                note_ids = []
```
RisingOrange: 🌱 This `find_notes` try/except isn't mentioned in the PR description and feels out of scope for NRT-645. It also swallows errors with a bare `except Exception` and logs a warning, then continues with `note_ids = []` — if a malformed search string is reaching this code, the right move is to understand where it originates and fix the source rather than silently treat the search as empty (which then opens a dialog with zero notes, leading to its own confusing UX). Suggest splitting into a separate PR or removing.

tests/addon/test_integration.py
```diff
@@ -9480,6 +9491,81 @@ def test_create_subdeck_screen_validation(
...
+            mocker.patch("aqt.mw.col.decks.children", return_value=[])
+
+            deck_config = config.deck_config(ah_did)
+            root_deck_id = deck_config.anki_id
+            dialog = BlockExamSubdeckDialog(root_deck_id, self.note_ids, parent=None)
+            qtbot.addWidget(dialog)
+
+            assert dialog.add_succeeded is False
+            dialog.reject()
+            assert dialog.add_succeeded is False
```
RisingOrange, line 9539: 🌱 The test name says "emits_cancelled_status_on_reject", but this test doesn't actually verify any postMessage emission — the `_on_js_message` `finished` callback (which is what posts `{status: "cancelled"}`) isn't wired up here. It only asserts that `add_succeeded` stays `False` on `reject()`. Either rename to reflect what it tests (the `add_succeeded` invariant on reject), or extend `test_add_to_block_exam_subdeck_pycmd` with a cancel case: set `dialog_instance.add_succeeded = False`, invoke the captured callback, and assert `post_msg_mock` was called with `status == "cancelled"`.

ankihub/gui/js_message_handling.py
ddevdan: Sorry it was from another changes I've made. Removing it.

RisingOrange (approved):

PR #1288, opened 2026-05-12
[NRT-749] Redesign Suggest-a-change dialog with "Include in suggestion" section

RisingOrange:
Replaces the silent "strip personally-protected fields from outgoing suggestions" behavior with an explicit per-field selector. Every edited field (change-note) or every non-empty field (new-note) plus every tag change surfaces in the dialog and the user picks what to include. Globally-protected fields are excluded from the picker at dialog-open time (so they never enter the user's allowlist); server-side enforcement is the backstop at submit.

## Screenshots

<table><tr>
<td>Redesigned Suggest-a-change dialog (AnKing change-note)</td>
<td>Non-AnKing new-note suggestion dialog with locked first-field tooltip</td>
</tr><tr>
<td align="center"><em>AnKing deck — change-note with Source widget</em></td>
<td align="center"><em>Non-AnKing deck — new-note flow; first-field checkbox is locked-checked (tooltip visible)</em></td>
</tr></table>

## Related issues

- [NRT-749](https://ankihub.atlassian.net/browse/NRT-749) (subtask of [NRT-508](https://ankihub.atlassian.net/browse/NRT-508))

## ⚠️ Rollout

Gated on the server feature flag `auto_protect_fields_when_edited` (shared with the auto-protect-on-edit work that already landed). Flip *after* that work has been deployed long enough for users to autosync once — the globally-protected-fields cache the dialog reads is seeded on install + every sync. The server still enforces protection on submit, so a cold cache only makes the dialog noisy, not unsafe.

## Data flow

Three phases for both single and bulk paths:

**1. Open (data shaping)**
- Entry point computes `compute_note_diffs(notes)` once, producing `Dict[NoteId, NoteDiff]` — edited fields, added/removed tags, AH-DB existence, media delta.
- Bulk path runs `any_suggestible_from_diffs` as an empty-state gate; if nothing's suggestible (incl. new-note candidates whose first field is globally-protected), a toast fires and the dialog never opens.

**2. Dialog (selection)**
- `SuggestionDialog` receives notes + diffs + the globally-protected-fields map.
- It snapshots prior per-(deck, note-type) deselections from `config.last_deselected_fields` and threads everything into `IncludeInSuggestionWidget`.
- The widget renders one section per note type for fields, plus separate Added/Removed Tags sections; the user toggles.

**3. Submit**
- On accept the dialog emits `SuggestionMetadata(change_type, comment, source, auto_accept, filters: BulkSuggestionFilters)` and persists the user's per-note-type deselections back to config.
- The closure that `open_suggestion_dialog_for_*` passed in as the dialog's `callback=` combines the dialog's filters with the original notes and calls `suggest_note_update` / `suggest_new_note` (single) or `suggest_notes_in_bulk` → `_suggestions_for_notes` (bulk).
- The per-note builders (`_change_note_suggestion` / `_new_note_suggestion`) emit one `NewNoteSuggestion` / `ChangeNoteSuggestion` per note via the per-note-type allowlist.

The diffs are widget-only — the submit-side builders re-derive equivalent data from each note. Threading `NoteDiff` through to the submit path is tracked in [NRT-771](https://ankihub.atlassian.net/browse/NRT-771) as part of the post-flag-flip cleanup.

## Proposed changes

> Dependency direction: the data-layer subsections (`main/suggestions.py` / `main/exporting.py` / `settings.py`) are additive with safe defaults; the GUI subsections consume them. Kept as one PR rather than split because everything is gated behind `AUTO_PROTECT_FEATURE_FLAG` (so it merges dark, no blast radius), and the data layer is only exercised by the GUI that lands with it — splitting would land a batch of functions nothing calls yet.

### "Include in suggestion" panel (`gui/suggestion_dialog.py`)

- New `IncludeInSuggestionWidget` in a bordered, palette-aware frame.
- Header: bold *"Include in suggestion"* + compact counter (`8/8`) + a *"Select which fields you want to submit changes for"* subtitle.
- One section per note type (Anki suffix `(<deck> / <user>)` stripped via the existing helper) plus separate **"Added Tags"** / **"Removed Tags"** sections.
- Section title sits on the section's select-all checkbox; clicking the title toggles the section. Long titles **dynamically elide** to fit the available width (full text on hover only when elided).
- Long tag names wrap to multiple lines; tag tooltips wrap as rich text.
- Tag names sort alphabetically; very long names show `…::leaf` with the full tag in the tooltip.
- Globally-protected fields are excluded from the picker by the widget at `_populate` time, so the user's resulting allowlist already omits them. Server enforces at submit; no client-side denylist re-application.

### Dialog-level UX

- Outer layout is `[content_row, button_box]` so the left panel stops at the bottom of the right column's content (above the buttons), not at the dialog's bottom edge. 2:3 column stretch; left column capped at 360px so a pathological note-type name can't push the dialog wide.
- Source label: "Source" → **"Source (Required)"**. For source `Other`: label → "Details", placeholder → "Describe the source and include relevant details...". Rationale placeholder → "Explain why this change should be made...".
- `QDialogButtonBox` gains `Cancel`. Post-submit *"No field or tag changes were detected"* dialog removed — in-dialog gating makes it unreachable; a tooltip fallback covers the rare sync-race case.
- On accept, the user's deselected fields are persisted per-(deck, note-type) via `DeckConfig.last_deselected_fields_per_note_type`. Re-checking a field clears its deselection; deselections persist across sessions even when the field isn't re-edited.

### Suggestion API (`main/suggestions.py`)

- New `PerNoteFilters(fields_to_include, tags_to_add, tags_to_remove)` and `BulkSuggestionFilters(fields_to_include_by_mid, tags_to_add, tags_to_remove)` carry the dialog's per-note-type selection through `suggest_note_update` / `suggest_new_note` / `suggest_notes_in_bulk` into the per-note builders. The widget excludes globally-protected fields at populate time, so they never enter the allowlist; the submit layer applies only the user's allowlist and leaves global protection to the server (matches main's pre-PR behavior at the submit layer).
- New `compute_note_diffs(notes) -> Dict[NoteId, NoteDiff]` runs the per-note diff against the local AnkiHub DB once per dialog-open (was three `to_note_data` calls per note previously) and is shared by the bulk-gate, the media-added check, and the widget.
- New `any_suggestible_from_diffs(notes, diffs, change_type, globally_protected_by_mid)` is the bulk-suggest empty-state gate; it mirrors `_suggestions_for_notes`'s classification (DELETE → existing-and-not-deleted AH notes; everything else → notes with at least one non-protected field edit or tag change, with new-note candidates also requiring the first field to be non-empty and non-protected). On miss, the entry point shows *"No changes to suggest. Try syncing with AnkiHub first."* and doesn't open the dialog.

### `to_note_data` opt-in (`main/exporting.py`)

- `to_note_data` / `_prepare_fields` gain `include_protected_fields=False`. The default keeps the legacy auto-strip behavior; suggestion paths set it to `True` when `AUTO_PROTECT_FEATURE_FLAG` is on, so the user picks what to send via the dialog instead of having fields silently dropped before the dialog ever sees them.

### `_Config` API (`settings.py`)

- `globally_protected_fields(ah_did) -> Dict[int, List[str]]` whole-map getter (defensive copy). Per-mid callers (`editor.py`, `browser.py`) do `.get(note.mid, [])` on the result; the dialog reads the whole map for widget construction.
- `set_last_deselected_fields` / `last_deselected_fields` for the deselection-persistence cache.

## How to reproduce

The dialog's **OK/submit button** stays disabled until **rationale + source (when required) are filled AND at least one field/tag is selected**. The "Source (Required)" widget only renders for notes in the AnKing deck — non-AnKing dialogs go straight from Change Type to Rationale.

1. Edit any field on an AnkiHub note → open suggestion dialog → confirm only the edited field appears under "Include in suggestion", pre-selected.
2. Bulk-suggest across notes from two note types → confirm one section per note type, only edited fields shown.
3. Add a new (non-AnkiHub) note in a deck whose note types are AnkiHub-managed → open suggestion dialog → confirm **all non-empty fields** appear (no diff baseline), tags appear in "Added Tags" only.
4. Edit a globally-protected field → confirm it's not listed in the dialog and not sent.
5. Bulk-suggest a set of notes with no local changes → toast fires, dialog doesn't open.
6. Same set with "Suggest to delete note" → dialog opens (DELETE doesn't need changes).
7. Click the section title text (not just its checkbox) → confirm it toggles the section.
8. Hover a long tag → tooltip wraps to multiple lines instead of one long line.
9. On an AnKing note, set Source = "Other" → label becomes "Details" with the new placeholder.
10. Click Cancel → dialog closes without applying changes.

## Tests

`TestFieldsToSuggestFilters` in `test_integration.py` covers the filter allowlists (single + bulk, change-note + new-note), deselection persistence + merge, flag-off legacy path, globally-protected exclusion + bulk-gate backstop, the bulk empty-state toast, and the end-to-end `config → dialog kwarg` seam. `TestSelectAllGroupController` in `test_unit.py` (qtbot) pins down `_SelectAllCheckBox` tri-state interactions.

## Out of scope (deferred)

- Editor toolbar Suggest-button gating ([NRT-751](https://ankihub.atlassian.net/browse/NRT-751)).

[NRT-749]: https://ankihub.atlassian.net/browse/NRT-749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NRT-508]: https://ankihub.atlassian.net/browse/NRT-508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NRT-751]: https://ankihub.atlassian.net/browse/NRT-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/browser/browser.py
+++ b/ankihub/gui/browser/browser.py
@@ -358,7 +358,7 @@ def _on_protect_fields_action(browser: Browser, nids: Sequence[NoteId]) -> None:
 
     # Note types are 1:1 with AH decks, so the note's mid uniquely determines the deck.
     ah_did = ankihub_db.ankihub_did_for_note_type(note.mid)
-    globally_protected = config.globally_protected_fields_for_note_type(ah_did, note.mid) if ah_did else []
+    globally_protected = config.globally_protected_fields(ah_did).get(note.mid, []) if ah_did else []
 
     new_fields_protected_by_tags = choose_subset(
         "Choose which fields of this note should be protected from updates.<br><br>"
--- a/ankihub/gui/decks_dialog.py
+++ b/ankihub/gui/decks_dialog.py
@@ -47,6 +47,7 @@ from ..main.note_type_management import (
     update_note_type_templates_and_styles,
 )
 from ..main.subdecks import SUBDECK_TAG, deck_contains_subdeck_tags
+from ..main.suggestions import AUTO_PROTECT_FEATURE_FLAG
 from ..main.utils import (
     all_notes_in_deck,
     get_deck_for_ah_did,
@@ -288,7 +289,7 @@ class DeckManagementDialog(QDialog):
 
         # Setup "Automatically protect fields when edited" (gated on a server-controlled flag
         # — the suggestion dialog needs to ship before this option is meaningful to users)
-        auto_protect_enabled = (config.get_feature_flags() or {}).get("auto_protect_fields_when_edited", False)
+        auto_protect_enabled = config.get_feature_flags().get(AUTO_PROTECT_FEATURE_FLAG, False)
         if auto_protect_enabled:
             self.box_auto_protect_fields = self._setup_box_auto_protect_fields_when_edited(selected_ah_did)
 
--- a/ankihub/gui/editor.py
+++ b/ankihub/gui/editor.py
@@ -16,6 +16,7 @@ from ..db import ankihub_db
 from ..db.models import AnkiHubNote
 from ..gui.menu import AnkiHubLogin
 from ..main.note_conversion import TAG_FOR_PROTECTING_ALL_FIELDS, protection_tag_for_field
+from ..main.suggestions import AUTO_PROTECT_FEATURE_FLAG
 from ..main.utils import is_tag_in_list, update_notes_with_named_undo
 from ..settings import (
     ANKI_INT_VERSION,
@@ -68,18 +69,13 @@ def _on_field_unfocus_auto_protect(changed: bool, note: Note, current_field_idx:
     """
     # Server-controlled rollout gate: the suggestion dialog needs to ship before
     # auto-protect is safe to enable (otherwise edits get silently stripped on suggest).
-    if not (config.get_feature_flags() or {}).get("auto_protect_fields_when_edited", False):
+    if not config.get_feature_flags().get(AUTO_PROTECT_FEATURE_FLAG, False):
         return changed
 
     ah_did = ankihub_db.ankihub_did_for_anki_nid(note.id)
-    if not ah_did:
+    if not ah_did or not config.deck_config(ah_did).auto_protect_fields_when_edited:
         return changed
 
-    deck_config = config.deck_config(ah_did)
-    if not deck_config.auto_protect_fields_when_edited:
-        return changed
-
-    # If all fields are already covered by the "All" tag, no per-field tag is needed.
     if is_tag_in_list(TAG_FOR_PROTECTING_ALL_FIELDS, note.tags):
         return changed
 
@@ -87,8 +83,7 @@ def _on_field_unfocus_auto_protect(changed: bool, note: Note, current_field_idx:
     if field_name == ANKIHUB_NOTE_TYPE_FIELD_NAME:
         return changed
 
-    # Skip if globally protected for this deck (managed by deck owner, not locally)
-    if field_name in config.globally_protected_fields_for_note_type(ah_did, note.mid):
+    if field_name in config.globally_protected_fields(ah_did).get(note.mid, []):
         return changed
 
     ah_note = AnkiHubNote.get_or_none(anki_note_id=note.id)
@@ -101,21 +96,23 @@ def _on_field_unfocus_auto_protect(changed: bool, note: Note, current_field_idx:
     if field_name.lower() not in ah_field_names_lower:
         return changed
 
-    # ah_note.fields omits empty fields, so treat a missing field name as empty
-    ah_field_value = (ah_note.fields or {}).get(field_name, "")
     protection_tag = protection_tag_for_field(field_name)
+    is_protected = is_tag_in_list(protection_tag, note.tags)
+    # ah_note.fields omits empty fields, so treat a missing field name as empty.
+    ah_field_value = (ah_note.fields or {}).get(field_name, "")
     should_be_protected = note[field_name] != ah_field_value
-    if should_be_protected and not is_tag_in_list(protection_tag, note.tags):
-        # Field differs from AnkiHub version — add protection if not already present
+    if is_protected == should_be_protected:
+        return changed
+
+    if should_be_protected:
         note.tags.append(protection_tag)
-        update_notes_with_named_undo(f"AnkiHub | Auto-protect {field_name}", note)
-        return True
-    if not should_be_protected and is_tag_in_list(protection_tag, note.tags):
-        # Field matches AnkiHub version — remove protection tag if present
-        note.tags = [t for t in note.tags if t.lower() != protection_tag.lower()]
-        update_notes_with_named_undo(f"AnkiHub | Auto-unprotect {field_name}", note)
-        return True
-    return changed
+        verb = "protect"
+    else:
+        protection_tag_lower = protection_tag.lower()
+        note.tags = [t for t in note.tags if t.lower() != protection_tag_lower]
+        verb = "unprotect"
+    update_notes_with_named_undo(f"AnkiHub | Auto-{verb} {field_name}", note)
+    return True
 
 
 def _on_suggestion_button_press(editor: Editor) -> None:
--- a/ankihub/gui/suggestion_dialog.py
+++ b/ankihub/gui/suggestion_dialog.py
@@ -661,6 +661,9 @@ class SuggestionDialog(QDialog):
     def _refresh_source_widget(self):
         if self._source_needed():
             self.source_widget.setup_for_change_type(change_type=self._change_type())
+            # DELETE's source (Duplicate Note) is optional, so don't claim "(Required)" there.
+            title = "Source (Required)" if self.source_widget.input_is_required() else "Source"
+            self.source_widget_group_box.setTitle(title)
             self.source_widget_group_box.show()
         else:
             self.source_widget_group_box.hide()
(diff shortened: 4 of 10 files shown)


cacoze:
Hey @RisingOrange one thing I noticed (not sure if it`s a bug, neither if it was introduced on this branch) but the source field stated to be required, but it`s not being required in fact when sugesting to delete notes.

https://github.com/user-attachments/assets/5a04eb08-a88b-4be5-9f72-d46e6990c9b0

RisingOrange:
> Hey @RisingOrange one thing I noticed (not sure if its a bug, neither if it was introduced on this branch) but the source field stated to be required, but its not being required in fact when sugesting to delete notes.

Good catch! Delete suggestions are the exception for which the source isn't required. Fixed it so that the label doesn't say the source is required for them

cacoze:
Two small things I noticed while reading through — not blocking, just curious what you think:

**1. "Globally-protected fields re-read at submit" — doc vs. code?**

The PR description says:
> The globally-protected-fields cache is re-read at submit (not just snapshotted at dialog open) so a sync racing with the user's interaction still strips fields that became protected mid-dialog.

But `IncludeInSuggestionWidget.suggestion_filters()` constructs `BulkSuggestionFilters` without `globally_protected_fields_by_mid`, and neither closure (`_on_suggestion_dialog_for_*_closed`) re-reads `config.globally_protected_fields(...)` before calling `suggest_*`. So `_apply_field_filters`'s `denylist` branch never engages at submit — protection is purely from the allowlist the widget produced at dialog-open time.

Functionally this is fine because the server enforces, and `ApplicationModal` blocks a same-thread sync from racing — but the dialog also drops back into the event loop while open (network calls, etc.), so it's not _impossible_ for the cache to change mid-dialog. Was the plan to drop that claim from the description, or to actually plumb `globally_protected_fields_by_mid=self._globally_protected` through `suggestion_filters()` so the denylist engages as a belt-and-suspenders?

```python
def suggestion_filters(self) -> BulkSuggestionFilters:
    return BulkSuggestionFilters(
        fields_to_include_by_mid=self.selected_field_names_by_mid(),
        tags_to_add=self.selected_tag_additions(),
        tags_to_remove=self.selected_tag_removals(),
        # globally_protected_fields_by_mid=...?  re-read here, or via the closure?
    )
```

**2. `compute_note_diffs` runs even when `AUTO_PROTECT_FEATURE_FLAG` is off (bulk path)**

In `open_suggestion_dialog_for_bulk_suggestion`, the diff pipeline runs unconditionally before the flag check:

```python
diffs = compute_note_diffs(notes)

if config.get_feature_flags().get(AUTO_PROTECT_FEATURE_FLAG, False) and not any_suggestible_from_diffs(...):
    ...
```

The result is still consumed for `added_new_media=` and `note_diffs=` plumbing, but with the flag off the widget never renders, so the only flag-off consumer is the `added_new_media` bool. The perf gate caps the worst case at ~2s for 500 notes so it's bounded, but it's a behavior change vs. main where the bulk path was cheaper pre-dialog.

Worth short-circuiting (cheaper `added_new_media` check + `note_diffs=None` when the flag is off), or are you treating the perf-gated ceiling as good enough until the flag flips and this becomes the only path?

Happy to defer either / both to NRT-771 if you'd rather keep this PR tight.

RisingOrange:
Thanks @cacoze!

**1. "Cache re-read at submit" — doc vs code**

> `IncludeInSuggestionWidget.suggestion_filters()` constructs `BulkSuggestionFilters` without `globally_protected_fields_by_mid`, and neither closure re-reads `config.globally_protected_fields(...)` before calling `suggest_*`. So `_apply_field_filters`'s `denylist` branch never engages at submit — protection is purely from the allowlist the widget produced at dialog-open time.

Your comment made me realize the simpler fix for the inconsistency is to drop the submit-side denylist for g.p. fields altogether, rather than make the description's claim accurate. Globally-protected fields are excluded from the list of shown fields when the dialog opens (in `IncludeInSuggestionWidget._populate`). That's all the client needs to do — the server enforces on submit regardless.

Cleaned up in 1473cbc1 (+ 6a92d321 for a stale docstring follow-up). Three pieces went: `BulkSuggestionFilters.globally_protected_fields_by_mid`, `PerNoteFilters.globally_protected_fields`, and the denylist branch in `_apply_field_filters` (now `_apply_field_allowlist`, mirroring `_apply_tag_allowlist`). The PR description was updated to match.

**2. `compute_note_diffs` running flag-off**

> The result is still consumed for `added_new_media=` and `note_diffs=` plumbing, but with the flag off the widget never renders, so the only flag-off consumer is the `added_new_media` bool. […] it's a behavior change vs. main where the bulk path was cheaper pre-dialog.

`compute_note_diffs` is replacing main's per-note `_added_new_media(note)` helper here, which did roughly the same per-note work un-batched, likely less. So the flag-off cost is at or under main's, not above it. 

I'd leave this one as-is.

cacoze (approved):

PR #1293, opened 2026-05-25
[NRT-773] Deck Management dialog QA fixes

RisingOrange:
## Related issues
- Tracked by [NRT-773].
- Follow-up to the Deck Management dialog redesign shipped in NRT-748 (the merged [#1259](https://github.com/AnkiHubSoftware/ankihub_addon/pull/1259)).

## Proposed changes
UI adjustments to the **Deck Management** dialog (`ankihub/gui/decks_dialog.py`), from QA + design feedback:

1. **Sections no longer "jump" when switching decks.** `_refresh_box_bottom_right` had two `addStretch()` calls (one between "Destination for New Cards" and "Note Types", one trailing). Two stretches split the leftover vertical space, so on decks with fewer sections the whole panel floated to a different vertical position. Replaced the middle stretch with `addSpacing(16)`, leaving a single trailing stretch → every section stacks from the top consistently across all decks.

2. **"Note Types" gets the same section spacing as "Destination for New Cards"** — a direct consequence of the change above (the 16px gap above it now matches the other section headers).

3. **"Learn about…" doc links hug their related control.** The three link `QLabel`s were built from triple-quoted multi-line HTML; `QLabel` renders the leading/trailing newlines as phantom blank lines (~2 extra line-heights above/below each link). Collapsed all three to single-line HTML, removing the excess space QA flagged.

4. **Note-type buttons sit closer together.** The three buttons used `addSpacing(4)`, but the parent `QVBoxLayout`'s default item spacing also applied on both sides of each spacer (~24px total gap on macOS). Moved them into their own layout with an explicit `setSpacing(4)`.

5. **Long deck names no longer scroll the deck list sideways.** The left deck list now elides long names with a trailing "…" (horizontal scrollbar off + `ElideRight`) and exposes the full name via a word-wrapped HTML tooltip on hover. The right-side deck title keeps wrapping onto a second line as before, which reads fine.

6. **Unsubscribe button color matches Figma.** The light-mode text color was a duller, more desaturated pink than the spec (read as "blurry/inactive"); bumped it to the Figma coral `#f38987`, with a matching softened `#f5a19f` for dark mode.

7. **Tooltip icon alignment on "Remove notes without review history…".** That checkbox/icon row was the only one missing a trailing `addStretch()`, so with its multi-line label the tooltip icon got pushed to the far right of the wide panel. Added the stretch to match the other rows. _(from PR review)_

8. **Tightened the "Note Types" header→buttons gap.** Removed an explicit `addSpacing(8)` between the label and the first button so it matches the other section headers (this is the label-to-button gap, separate from the button-to-button tightening in #4). _(from PR review)_

No behavioral changes — purely layout/markup/styling. (Also merged latest `main`, which added auto-refresh-on-focus to this dialog; the long-name tooltip is applied in the new `_populate_decks_list` so it covers the auto-refresh path too.)

## How to reproduce
1. Open **AnkiHub → Deck Management**.
2. Select a subscribed deck you own (so "Note Types" + the auto-protect option render; the latter requires the `auto_protect_fields_when_edited` feature flag).
3. Observe: doc links sit directly under their controls, "Note Types" has consistent spacing above and below its header, the publish buttons are tightly grouped, the "Remove notes…" tooltip icon sits next to its checkbox, and the Unsubscribe button is the brighter coral.
4. Switch between decks — sections stay anchored at the same vertical positions (no jump).
5. Subscribe to a deck with a very long name — the list elides it with "…" (full name in the tooltip) instead of scrolling sideways.

## Screenshots and videos
Long deck name: elided in the list with a word-wrapped tooltip, and wrapped in the right-side title.

Deck Management: long deck name elided in the list with a wrapped tooltip, and wrapped in the right-side title

## Further comments
Stacked conceptually on the merged NRT-748 redesign; branched off `main`.

[NRT-748]: https://ankihub.atlassian.net/browse/NRT-748
[NRT-773]: https://ankihub.atlassian.net/browse/NRT-773

### DIFF
--- a/ankihub/gui/decks_dialog.py
+++ b/ankihub/gui/decks_dialog.py
@@ -2,6 +2,7 @@
 
 import uuid
 from concurrent.futures import Future
+from html import escape
 from typing import List, Optional
 from uuid import UUID
 
@@ -185,6 +186,10 @@ class DeckManagementDialog(QDialog):
         self.decks_list_label.setSizePolicy(QSizePolicy.Policy.Expanding, QSizePolicy.Policy.Preferred)
 
         self.decks_list = QListWidget()
+        # Elide long deck names with a trailing "…" instead of showing a horizontal
+        # scrollbar; the full name is available via each item's tooltip.
+        self.decks_list.setHorizontalScrollBarPolicy(Qt.ScrollBarPolicy.ScrollBarAlwaysOff)
+        self.decks_list.setTextElideMode(Qt.TextElideMode.ElideRight)
         qconnect(self.decks_list.itemSelectionChanged, self._refresh_box_bottom_right)
 
         box = QVBoxLayout()
@@ -883,6 +888,10 @@ class DeckManagementDialog(QDialog):
                 item = QListWidgetItem(f"{deck.name} (Maintained by you)")
             else:
                 item = QListWidgetItem(deck.name)
+            # Full name on hover, since long names are elided in the list. Wrap it in
+            # HTML so Qt renders the tooltip as a word-wrapped block instead of one long
+            # thin line; escape the name so any "<"/"&" in it isn't treated as markup.
+            item.setToolTip(f"<div>{escape(item.text())}</div>")
             item.setData(Qt.ItemDataRole.UserRole, deck)
             self.decks_list.addItem(item)
 
--- a/ankihub/gui/decks_dialog.py
+++ b/ankihub/gui/decks_dialog.py
@@ -543,10 +543,13 @@ class DeckManagementDialog(QDialog):
             ),
         )
 
-        # Add checkbox and icon label to the result layout
+        # Add checkbox and icon label to the result layout. The trailing stretch keeps
+        # the icon next to the (multi-line) checkbox instead of being pushed to the far
+        # right of the wide panel, matching the other checkbox/icon rows.
         box = QHBoxLayout()
         box.addWidget(self.ankihub_deleted_notes_behavior_cb)
         box.addWidget(self.ankihub_deleted_notes_behavior_icon_label)
+        box.addStretch()
 
         return box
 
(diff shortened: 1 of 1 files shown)


cacoze:
I've spotted two possible layout issues when testing the branch locally that could be worth fixing while we're at it:

image

Here are the problems and some suggested fixes:

### 1. Tooltip icon pushed far right on "Remove notes without review history…"

In `_setup_box_ankihub_deleted_notes_behavior`, the checkbox + icon row is missing the trailing `addStretch()` that every other checkbox/icon row in this dialog uses (e.g. suspend-new-cards, subdecks, auto-protect). Combined with the multi-line `QCheckBox` label, the checkbox expands horizontally in the wide right panel and the icon gets pushed to the far right.

```python
box = QHBoxLayout()
box.addWidget(self.ankihub_deleted_notes_behavior_cb)
box.addWidget(self.ankihub_deleted_notes_behavior_icon_label)
# missing: box.addStretch()
```

**Suggested fix:** add `box.addStretch()` after the icon, matching the other rows.

---

### 2. Extra space between "Note Types" header and "Publish note type"

In `_setup_box_note_types`, there's an explicit `addSpacing(8)` between the section label and the buttons. Other section headers (e.g. "Deck Options", "Destination for New Cards") go straight from label → content with no fixed spacer. On top of that, the parent `QVBoxLayout`'s default item spacing applies on both sides of the 8px spacer (~6–9px on macOS), so the effective gap is closer to ~20px.

The nested `buttons_layout` from this PR correctly tightens spacing *between* the three buttons — it doesn't affect the label-to-first-button gap.

**Suggested fix:** drop (or reduce) the `addSpacing(8)` after `note_types_label` so it matches the other sections.

cacoze:
Sorry for the push-force.  I was trying to submit the attached image through cursor agent and I've let him go through a bad path by doing unnecessary commits. I push forced to remove those commits.

RisingOrange:
> I've spotted two possible layout issues when testing the branch locally that could be worth fixing while we're at it

Good finds! Fixed them

cacoze (approved):

PR #1294, opened 2026-05-26
[NRT-763] Auto-refresh Deck Management dialog when it regains focus

RisingOrange:
## Related issues
- Closes [NRT-763](https://ankihub.atlassian.net/browse/NRT-763)

## Proposed changes
The **Deck Management** dialog only refreshed its subscription list on open, unsubscribe, and reopen. So when a user clicked **🔗 Browse Decks**, subscribed to a deck on the web, and switched back to the still-open dialog, the new deck didn't appear until they closed and reopened the dialog.

This makes the dialog auto-refresh when it regains focus. Returning from the external browser raises a window-activation event (not a focus-in/show), so the trigger is `changeEvent()` + `QEvent.Type.ActivationChange`. It only acts on a **genuine return** — the window was deactivated and then reactivated — which naturally excludes the activation Qt emits while the dialog is first shown (no timing heuristic needed).

The refresh is built to be cheap and non-disruptive:

- **Async fetch** via `AddonQueryOp` (no progress spinner) — `get_deck_subscriptions()` is a blocking HTTP call, so it never runs on the UI thread.
- **Diff before render:** an order-insensitive `(ah_did, name, user_relation)` snapshot is compared against the current list; the list is only rebuilt when it actually changed. Unchanged returns are no-ops → no flicker, current selection preserved.
- **Selection on change:** if nothing is currently selected and one or more decks were added (the common "just subscribed on the web" case), one is auto-selected and scrolled into view, surfacing its panel — including the "⚠️ not installed → 🔃 Sync to install" nudge. When several decks were added at once, the one appearing last in the server response is picked (there is no per-deck "subscribed at" timestamp, so this is the closest proxy for "last added"). If the user already has a deck selected, that selection is kept, so a refresh can't retarget an in-progress, deck-scoped workflow (publish / change-destination).
- **Debounce + defer:** a 400 ms debounce coalesces bursts of activations into one refresh. If a fetch is already in flight, the return is deferred (the timer re-arms) rather than dropped, so a change the in-flight request didn't capture is still picked up.
- **Safety:** the auto path is skipped while a child modal/popup is open (closing it reactivates the dialog, re-triggering); async results are versioned so a stale in-flight fetch can't clobber a newer (e.g. post-unsubscribe) refresh; callbacks are guarded with `sip.isdeleted(self)` / `isVisible()` (the dialog is retained by a singleton, so it can be closed-but-not-deleted); and the rebuild blocks signals to avoid the `itemSelectionChanged → _refresh_box_bottom_right` cascade flashing the right panel.
- **Errors:** failures on the auto path are logged silently — no error popup on every focus when offline.

All changes are contained to `ankihub/gui/decks_dialog.py`.

## How to reproduce
1. AnkiHub menu → **Deck Management**.
2. Click **🔗 Browse Decks** — the deck catalog opens in the browser.
3. Subscribe to a deck on the web.
4. Switch back to the still-open Deck Management dialog.
5. **Before:** the new deck is missing until you close and reopen the dialog. **After:** the deck appears (and, if nothing was selected, is selected) shortly after the dialog regains focus.

## Screenshots and videos
_No UI screenshots — behavioural change._ Verified manually on Linux (subscribe-on-web → return → deck appears on focus; no flicker on unchanged returns) in addition to the automated tests below.

## Further comments
A manual "Refresh" button was considered as a simpler fallback; auto-refresh-on-focus was chosen so it Just Works, with the diff/defer/guards above keeping it safe. Added tests (`test_auto_refresh_*`) covering: auto-select of a single new deck, auto-select of the last-in-response deck when several are added at once, keeping an existing selection when a deck is added, no-op on unchanged subscriptions, deferral while a fetch is in flight, the stale-async-result generation guard, the child-modal/popup guard, and that a refresh only triggers on a genuine return (not the initial show).

[NRT-763]: https://ankihub.atlassian.net/browse/NRT-763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/decks_dialog.py
+++ b/ankihub/gui/decks_dialog.py
@@ -1037,12 +1037,16 @@ class DeckManagementDialog(QDialog):
         added_ids = new_ids - old_ids
 
         current = self._selected_ah_did()
-        if current is None and len(added_ids) == 1:
-            # Nothing selected and exactly one new subscription (the common "just
-            # subscribed on the web" case): select it so its panel — including
-            # "Sync to install" — is shown. If the user already has a deck selected
-            # we keep it, so a refresh can't retarget an in-progress workflow.
-            select_ah_did: Optional[UUID] = next(iter(added_ids))
+        if current is None and added_ids:
+            # Nothing selected and at least one new subscription (the common "just
+            # subscribed on the web" case): auto-select one so its panel — including
+            # "Sync to install" — is shown. When the user subscribed to several decks
+            # before returning, pick the one that appears last in the server response
+            # (we have no per-deck "subscribed_at" timestamp; this is the closest
+            # proxy for "last added" and matches the order rendered in the list).
+            # If the user already has a deck selected we keep it, so a refresh can't
+            # retarget an in-progress workflow.
+            select_ah_did: Optional[UUID] = next(deck.ah_did for deck in reversed(decks) if deck.ah_did in added_ids)
         else:
             select_ah_did = current if current in new_ids else None
 
--- a/ankihub/gui/decks_dialog.py
+++ b/ankihub/gui/decks_dialog.py
@@ -102,8 +102,10 @@ class DeckManagementDialog(QDialog):
         # genuine return (deactivate -> reactivate), not the activation Qt emits
         # while the dialog is first shown.
         self._was_deactivated = False
-        # Bumped on every fetch (sync or async); an async result whose generation is
-        # stale (a newer fetch/refresh started meanwhile) is discarded on completion.
+        # Prevents a slow async auto-refresh from clobbering a newer refresh with
+        # stale data (e.g. an unsubscribe rebuilt the list while the fetch was still
+        # in flight). Every fetch — sync or async — bumps this counter, and an async
+        # result is only applied if its generation still matches on completion.
         self._subscriptions_fetch_generation = 0
         # Debounces focus-triggered refreshes; parented to self so it can't fire
         # after the dialog is deleted.
--- a/ankihub/gui/decks_dialog.py
+++ b/ankihub/gui/decks_dialog.py
@@ -1,8 +1,9 @@
 """Dialog for managing subscriptions to AnkiHub decks and deck-specific settings."""
 
+import time
 import uuid
 from concurrent.futures import Future
-from typing import List, Optional
+from typing import List, Optional, Tuple
 from uuid import UUID
 
 import aqt
@@ -15,6 +16,7 @@ from aqt.qt import (
     QComboBox,
     QDialog,
     QDialogButtonBox,
+    QEvent,
     QFrame,
     QHBoxLayout,
     QLabel,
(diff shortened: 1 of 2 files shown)


cacoze:
LGTM!

Approving, but pointing one nice-to-have for follow-up — two test cases that would lock in branches the current suite doesn't exercise directly. Both are minor (not blocking this PR), but they cover load-bearing guards that a future cleanup could silently regress:

1. **Generation discard.** A test that pins the invariant: if a newer sync refresh (e.g. unsubscribe) runs while an async auto-refresh is in flight, the stale async result must be dropped on completion and must not clobber the newer list. The shape would be: patch `AddonQueryOp` to capture the `on_success` callback without firing it, call `_auto_refresh_decks_list()`, then `_refresh_decks_list()` (bumps the generation), then fire the captured `on_success` with a stale list and assert `_apply_fetched_subscriptions` was not called and the list count is unchanged.

2. **Modal-open guard.** The existing `test_auto_refresh_defers_while_fetch_in_flight` deliberately mocks `activeModalWidget`/`activePopupWidget` to `None` to isolate the in-flight branch, so the guard itself is untested. Two small tests would cover it: (a) patch `activeModalWidget` to return a child `QDialog(dialog)` and assert `_on_refresh_debounce_timeout()` neither fetches nor re-arms (it's recovered later by the ActivationChange when the child closes); (b) patch `activeModalWidget` to return `dialog` itself and assert the refresh still proceeds — this is the `is not self` carve-out that keeps auto-refresh working at all when the dialog is application-modal. A one-liner variant for `activePopupWidget` (combo dropdown open) is worth adding too.

Both branches are short conditionals next to other conditionals, so they're exactly the kind of thing a future "clean up these guards" refactor could drop without anything turning red. Happy to leave it to you whether to add them in this PR or in a follow-up.

cacoze (dismissed):

RisingOrange:
@cacoze Good suggestions, added the tests

cacoze (dismissed):

RisingOrange:
@cacoze I added a change to address this [QA comment](https://ankihub.atlassian.net/browse/NRT-763?focusedCommentId=43984):
> One thing that I noticed if the user subscribe in more than one deck, the newly added deck isn’t auto-selected. In that case we should auto-selected the last added.

Commit: [[NRT-763] Auto-select last newly-added deck when several appear at once](https://github.com/AnkiHubSoftware/ankihub_addon/pull/1294/commits/b7bc7040846e4bdf218d35575d529566603bbc65)

[NRT-763]: https://ankihub.atlassian.net/browse/NRT-763?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

cacoze (approved):

PR #1295, opened 2026-05-26
[NRT-764] Fix macOS modal deadlock: one captured parent for sync-flow dialogs

RisingOrange:
## Related issues
- [NRT-764](https://ankihub.atlassian.net/browse/NRT-764) — "AnkiHub | Sync" install sheet appears behind Deck Management dialog on macOS → modal deadlock

image

## Proposed changes
On macOS, clicking **Sync to install** in the Deck Management dialog caused a hard modal deadlock: the "AnkiHub | Sync" confirmation sheet (and the sync flow's progress popups) were parented to the main Anki window (`aqt.mw`). Qt renders a dialog parented to the main window as a **sheet attached to it**, z-ordered *behind* the `ApplicationModal` Deck Management dialog. The hidden sheet couldn't be interacted with, and Deck Management couldn't be closed — only `kill -9` recovered.

This mirrors the existing **Unsubscribe** dialog, which layers correctly because it's parented to the Deck Management dialog itself rather than to `aqt.mw`. So the deadlock is fundamentally a wrong-parent bug, not a problem with layering.

### Approach: one captured parent for the whole sync flow
The sync flow is asynchronous and multi-step, so resolving the parent independently at each dialog/progress construction point is fragile: a popup created while a just-closed `ProgressDialog` is still the active window (or before focus returns to Deck Management) would fall back to `aqt.mw` and reintroduce the deadlock.

Instead, a new helper module **`ankihub/gui/sync_dialog_parent.py`** captures the parent **once**, at the start of `sync_with_ankihub()` — when the launching dialog (e.g. Deck Management) is reliably the active window and no progress popup exists yet — and every dialog/progress in the flow reuses it via `sync_dialog_parent()`. The helper guards the captured widget with `sip.isdeleted` + `isVisible` (Deck Management is a hidden-on-close singleton) and falls back to `active_window_or_mw()` when there's no usable captured parent.

This keeps function signatures unchanged (the shared `sync_with_ankihub` / `download_and_install_decks` entry points and their other callers are untouched) and reuses the codebase's existing single-flight sync-state idiom.

Dialogs/progress now routed through the captured parent:
- The "AnkiHub | Sync" confirmation dialog (`new_deck_subscriptions.py`), set **`WindowModal`** so it's interactive *and* renders as a sheet attached to its parent on macOS.
- The 3 sync `AddonQueryOp` progress popups and the `ChangesRequireFullSyncDialog` (`ankihub_sync.py`).
- Both deck-download `with_progress` popups and the import-summary dialog (`deck_installation.py`), the latter also `WindowModal` for the same sheet-attachment reason.
- The post-install `clear_unused_tags` cleanup progress (`deck_installation.py`).
- The subdeck due-date reminder (`subdeck_due_date_dialog.py`).
- The post-sync DB-check prompts (`db_check/*.py`) and deck-update error dialogs (`deck_updater.py`). The deck-update 403 handler runs on a background thread, so its parent is resolved inside `run_on_main`.

When sync is triggered from anywhere other than a dialog, `sync_dialog_parent()` resolves to `aqt.mw` — identical to previous behavior — so non-macOS / non-Deck-Management flows are unchanged. No `.exec()` calls were added; the existing non-blocking `.open()`/`.show()`/callback pattern is preserved. On success/cancel/fail, focus returns to Deck Management, which keeps its selection (the deck list is not rebuilt in this path).

### Out of scope
- Anki's own **AnkiWeb full-sync UI** (`aqt.sync.sync_collection`, parented to `mw`) can still appear behind Deck Management on macOS. It's Anki-native and can't be cleanly reparented; it belongs to the broader "convert macOS sheets / agree a modality standard" sweep the ticket already calls out as a separate pass.

## How to reproduce
**Before (bug), on macOS:**
1. Open Deck Management.
2. Select a subscribed-but-not-installed deck → right pane shows "This deck is not installed yet!" with a **Sync to install** button.
3. Click **Sync to install**.
4. The "AnkiHub | Sync" sheet appears behind the Deck Management dialog; clicking it does nothing, and Deck Management can't be closed → Anki must be force-killed.

**After (fixed):** the "AnkiHub | Sync" dialog (and every subsequent sync/install dialog and progress popup) layers above Deck Management as a sheet and is interactive; on completion/cancel/failure it closes and focus returns to Deck Management with the same deck still selected.

Linux/Windows render free-floating dialogs, so the deadlock never manifested there; this change is behavior-neutral on those platforms.

## Screenshots and videos
Verified on macOS (Apple Silicon).

**Sync triggered from Deck Management** — the "AnkiHub | Sync" confirmation dialog now slides down as a sheet attached to Deck Management instead of hiding behind it:

Sync confirmation dialog as a sheet attached to Deck Management on macOS

After **Install**, the "AnkiHub | Deck Import Summary" also attaches as a sheet to the same Deck Management dialog (left), and behaves identically as a sheet on the main window when sync is triggered without Deck Management open (right):

<table><tr>
<td>Deck Import Summary as a sheet attached to Deck Management on macOS</td>
<td>Deck Import Summary as a sheet attached to the main Anki window on macOS</td>
</tr></table>

## Further comments
- Alternatives considered: hiding/closing Deck Management during sync (rejected by design — they want the layered pattern, consistent with Unsubscribe); threading an explicit `parent` argument through the whole async chain (rejected — heavy signature churn through callback-heavy code and shared entry points, for what is UI context rather than business data). Both Codex (gpt-5.5) and a local review pass independently recommended the captured-parent approach.
- Testing: `ruff` + `mypy` clean across all changed files; 81 integration tests pass (confirmation dialog, sync, auto-sync, install, subdeck reminders, DB-check, deck-updater). Manually verified on macOS (Apple Silicon).

[NRT-764]: https://ankihub.atlassian.net/browse/NRT-764?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


RisingOrange:
Superseded by #1298. This PR took the layered approach (parent the sync dialogs to Deck Management as macOS sheets). macOS testing showed that's fundamentally fragile for the install flow: Anki's own AnkiWeb sync UI is parented to the main window and can't be re-layered, and the flow runs a full-upload sync *after* the import summary, which tangles with the summary's layering. #1298 takes the simpler, robust route — close Deck Management during install and return via the summary's existing "Go to Deck Management" button.

PR #1297, opened 2026-05-26
[NRT-751] Disable the editor Suggest button when there's nothing to suggest

RisingOrange:
## Related issues

- [NRT-751](https://ankihub.atlassian.net/browse/NRT-751) — sub-task of [NRT-508](https://ankihub.atlassian.net/browse/NRT-508) ("Automatically protect a field when users update it"), deferred from [NRT-749](https://ankihub.atlassian.net/browse/NRT-749).
- Based on #1288 (`NRT-749`), whose diff-based suggestion API (`compute_note_diffs` / `any_suggestible_from_diffs`) this reuses. **Merge #1288 first; this PR targets `NRT-749`.**

## Proposed changes

When a user is in the editor for an AnkiHub note that has nothing to suggest (no field diff vs the AnkiHub-stored note, no tag change), the toolbar's "Suggest a change" button is now disabled with the tooltip **"No changes to suggest"**. Previously the button was always enabled and submitting opened a dialog that couldn't produce a suggestion. Implements the NRT-508 spec line: *"If the user didn't edit any field the 'Suggest a change' button should be disable[d]."*

Gating lives in `ankihub/gui/editor.py`, behind the `auto_protect_fields_when_edited` feature flag (matching NRT-749's dark-merge strategy):

- **Visual gate** — in `_refresh_buttons`, both the change-note (not-deleted) and new-note branches disable `SUGGESTION_BTN_ID` and set a tooltip explaining why when the note has nothing to suggest: *"No changes to suggest"*, or *"The first field can't be empty"* (the latter is a new note's only possible gate reason). View / View-history buttons are managed per branch as before.
- **Action gate** — in `_on_suggestion_button_press`, an early-return covers the keyboard-shortcut path (the shortcut fires even when the button is visually disabled): a no-op for non-AnkiHub note types (closing a pre-existing `AssertionError` from the dialog opener), the gate tooltip when nothing's suggestible (including empty/globally-protected-first-field **new** notes, so they don't fall through to `add_current_note`), and the deleted-note tooltip for notes deleted on AnkiHub.
- **Live refresh** — registered on `editor_did_fire_typing_timer` (field edits) and `editor_did_update_tags` (tag edits) so the button tracks edits within the current note instead of only updating on note reload.

"Suggestible" is computed via `any_suggestible_from_diffs` — the same check the bulk-suggest dialog uses — so the button and the dialog agree: a field edit outside the **globally-protected** denylist, or a tag change (and, for a new note, a non-empty first field). Locally/personally-protected fields (`AnkiHub_Protect::Field`) still count: the protect tag only stops sync from overwriting the local edit; the user can still suggest it. Only globally-protected (deck-owner-managed) fields are excluded.

The gate predicate returns the tooltip reason (or `None` when suggestible) rather than a bare bool, so each block reason shows the right message. For a new note (no AnkiHub-DB row) the deck is resolved from its note type, so a globally-protected first field gates the button just as it would block the dialog. The button's enabled-state tooltip was also rephrased from *"Send your request to AnkiHub"* to *"Suggest a change to AnkiHub"* (or *"Suggest the note to AnkiHub"* for new notes) — "request" was the odd word out vs the feature's "Suggest" vocabulary. The button label/tooltip JS setters now encode their value with `json.dumps` — a latent bug surfaced by the apostrophe in the new tooltip (values were interpolated into single-quoted JS literals with no escaping).

### Editor-registry, and the removed note-type handler

The live-refresh hooks pass only a `Note`, not the `Editor`, so this PR adds a `_tracked_editors` registry (populated on `editor_did_init`, pruned lazily, matched by object identity) used by the field-typing and tag-update handlers.

While here, the pre-existing `editor` module global and its `add_cards_did_change_note_type` handler were removed: a note-type change in AddCards goes through `editor.loadNote`, which already fires `editor_did_load_note -> _refresh_buttons` on the affected editor (verified in aqt at 2.1.50 and 25.x). The explicit handler was redundant and actually fired *earlier* — synchronously after `loadNote()` returns, before the webview callback — racing the DOM.

### Note: delete-from-editor for unchanged notes

The gate treats `change_type` as non-DELETE: an unchanged existing AnkiHub note shows *"No changes to suggest"* and the button is disabled, even though the dialog's change-type dropdown *would* let the user pick `DELETE` (`suggest_note_update` exempts DELETE from its empty/no-changes checks). This matches the NRT-508/751 spec wording — *"if the user didn't edit any field the button should be disabled"* — and removes one minor pre-existing path: deleting unchanged notes via the editor's Suggest button. The browser's **"Suggest to delete"** menu remains the supported path for that flow.

## How to reproduce

With the `auto_protect_fields_when_edited` flag on:

1. Open an AnkiHub note in the editor with no local edits → Suggest button is disabled, tooltip "No changes to suggest".
2. Edit a field → button enables live, without reloading the note. Revert it → disables again.
3. Edit only a globally-protected field → stays disabled. A locally-protected field edit keeps it enabled.
4. Add or remove a tag → button updates live.
5. Add a new note of an AnkiHub note type with an empty first field → button is disabled; type into the first field → it enables.
6. With no changes, trigger the suggest hotkey → tooltip fires, dialog does not open. Same for a note deleted on AnkiHub.

With the flag off, the button behaves as before (always enabled for valid notes).

## Tests

New tests in `TestEditor` (`tests/addon/test_integration.py`), consolidated into three parametrized tests + four standalone ones: the existing-note visual gate (no-change disabled, field edit enabled, globally-protected-only edit excluded, locally-protected edit still enabled, flag-off no-op), the new-note visual gate (empty / filled / globally-protected-first-field), the action-gate blocks (no-changes, deleted, empty-new-note keyboard paths), live refresh on typing-timer + tag-update (existing notes) and on typing into a new note, and a no-op test for the non-AnkiHub hotkey path (logged-in / logged-out). All 26 `TestEditor` tests pass; mypy + ruff clean.

[NRT-751]: https://ankihub.atlassian.net/browse/NRT-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NRT-749]: https://ankihub.atlassian.net/browse/NRT-749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[NRT-508]: https://ankihub.atlassian.net/browse/NRT-508

### DIFF
--- a/ankihub/gui/editor.py
+++ b/ankihub/gui/editor.py
@@ -50,7 +50,8 @@ def _setup_additional_editor_buttons():
     gui_hooks.webview_did_receive_js_message.append(_on_js_message)
     gui_hooks.editor_did_load_note.append(_refresh_buttons)
     gui_hooks.editor_did_init.append(_track_editor)
-    gui_hooks.editor_did_fire_typing_timer.append(_on_editor_typing_timer)
+    gui_hooks.editor_did_fire_typing_timer.append(_refresh_buttons_for_edited_note)
+    gui_hooks.editor_did_update_tags.append(_refresh_buttons_for_edited_note)
     gui_hooks.add_cards_did_change_note_type.append(_on_add_cards_did_change_notetype)
 
     Editor.ankihub_command = AnkiHubCommands.CHANGE.value  # type: ignore
@@ -439,8 +440,9 @@ def _set_suggestion_button_tooltip(editor: Editor, text: str) -> None:
 
 editor: Editor
 
-# Live editors tracked so the typing-timer hook (which only receives a Note) can
-# locate the owning editor and refresh its button state as the user types.
+# Live editors tracked so the field-typing and tag-update hooks (which only
+# receive a Note) can locate the owning editor and refresh its button state as
+# the user edits.
 _tracked_editors: List[Editor] = []
 
 
@@ -448,7 +450,12 @@ def _track_editor(editor: Editor) -> None:
     _tracked_editors.append(editor)
 
 
-def _on_editor_typing_timer(note: Note) -> None:
+def _refresh_buttons_for_edited_note(note: Note) -> None:
+    """Refresh the owning editor's buttons after a field or tag edit.
+
+    Hook handler for editor_did_fire_typing_timer and editor_did_update_tags,
+    both of which pass only the edited Note.
+    """
     # Match by object identity, not note.id: unsaved Add-Cards notes all share
     # id 0, so an id comparison would refresh every open Add-Cards editor.
     for tracked in list(_tracked_editors):
--- a/ankihub/gui/editor.py
+++ b/ankihub/gui/editor.py
@@ -152,7 +152,7 @@ def _on_suggestion_button_press(editor: Editor) -> None:
     # Cover the rest of the keyboard-shortcut path: it fires even when the
     # button is visually disabled, including for empty new notes that must
     # not fall through to add_current_note below.
-    ah_note = AnkiHubNote.get_or_none(anki_note_id=note.id) if note.id != 0 else None
+    ah_note = AnkiHubNote.get_or_none(anki_note_id=note.id)
     if ah_note is not None and cast(AnkiHubNote, ah_note).was_deleted():
         tooltip(NOTE_DELETED_TOOLTIP)
         return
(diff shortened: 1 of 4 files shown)


RisingOrange:
Closing in favor of https://github.com/AnkiHubSoftware/ankihub_addon/pull/1304

PR #1303, opened 2026-06-01
[NRT-766] Rename ID Pending/With AnkiHub ID labels and fix Not on AnkiHub filter

RisingOrange:
## Related issues
- [NRT-766](https://ankihub.atlassian.net/browse/NRT-766) — Update "ID Pending" and "With AnkiHub ID" Labels

## Proposed changes
The "ID Pending" sidebar category under the AnkiHub section has caused recurring community confusion (users don't know if it's a bug, a queue, or something to act on). This PR addresses the ticket's three parts:

**1. Filter fix** — the "Not on AnkiHub" entry previously used an `ankihub_id:` field-name search, which matched any note whose note type merely *carries* an `ankihub_id` field (e.g. cloned / import-collision note-type variants). Those notes can't actually be suggested, so every action was greyed out — the source of the confusion described in the [Discourse thread](https://community.ankihub.net/t/cards-on-id-pending/596332).

The filter now gates on the note type being **registered with an AnkiHub deck** (new `AnkiHubNoteTypeSearchNode`, which uses the same `is_ankihub_note_type` check the suggestion entrypoints already enforce). After this change, "Not on AnkiHub" contains exactly the notes a user can act on. The filter is `ankihub_note_type:yes AND ankihub_note:no`.

**2. Label rename** ("everywhere in the product"):
- `With AnkiHub ID` → **On AnkiHub**
- `ID Pending` → **Not on AnkiHub**

Applied to the browser sidebar items, the two bulk-suggest info messages that reference those sidebar entries, and the "AnkiHub ID" browser column's empty-field display value.

**3. Tooltip** — the "Not on AnkiHub" item gets a trailing **ⓘ** glyph and an interactive tooltip explaining why notes appear there, with a clickable **suggested** link to the community guide (opens in the browser). Anki's built-in sidebar tooltip (`ToolTipRole`) can't host a clickable link, so this is a small frameless popup (`ankihub/gui/browser/sidebar_tooltip.py`) attached non-invasively via the `browser_will_show` hook — no custom item delegate, so Anki's sidebar rendering is untouched. It's built to feel native:
- Matches Anki's tooltip colors (`CANVAS` / `FG`), uses the platform tooltip wake-up delay, and draws a crisp uniform 1px border (DPI-aware; thin on macOS, themed gray/FG elsewhere).
- Opens/closes via cursor-position polling rather than Enter/Leave events, which is reliable across the row→popup gap on macOS (where a non-activating popup doesn't get mouse-move events).
- Theme-aware on both light and dark (the link color uses `FG_LINK`).

> One macOS limitation we accepted: the link doesn't switch to a hand cursor on hover, because macOS won't apply a cursor over a non-activating (non-key) window. The link is still styled and clickable.

Out of scope (the ticket says so): the wrong-note-type UX (disabled-button tooltips, a guided "change to the AnkiHub note type" action) is tracked separately.

## How to reproduce
1. Install/sync an AnkiHub deck.
2. Open the Anki Browser → expand **👑 AnkiHub** in the left sidebar.
3. The categories now read **On AnkiHub** and **Not on AnkiHub ⓘ**.
4. Hover over **Not on AnkiHub ⓘ** → the tooltip explains why notes appear there; click **suggested** to open the community guide.
5. Create a new note using an AnkiHub note type but don't suggest it → it shows under **Not on AnkiHub** and can be suggested. A note on a cloned/derivative (unregistered) note-type variant no longer appears there.

## Screenshots and videos
The **Not on AnkiHub ⓘ** sidebar item with its interactive tooltip, in dark mode — Linux (left) and macOS (right):

<table><tr>
<td>Tooltip in dark mode on Linux</td>
<td>Tooltip in dark mode on macOS</td>
</tr></table>

[NRT-766]: https://ankihub.atlassian.net/browse/NRT-766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/db/__init__.py
+++ b/ankihub/db/__init__.py
@@ -1,6 +1,8 @@
 from .db import (  # noqa: F401
+    DEFAULT_CHUNK_SIZE,
     NOTE_NOT_DELETED_CONDITION,
     ankihub_db,
+    chunks,
     execute_list_query_in_chunks,
     flat,
 )
--- a/ankihub/db/__init__.py
+++ b/ankihub/db/__init__.py
@@ -1,8 +1,6 @@
 from .db import (  # noqa: F401
-    DEFAULT_CHUNK_SIZE,
     NOTE_NOT_DELETED_CONDITION,
     ankihub_db,
-    chunks,
     execute_list_query_in_chunks,
     flat,
 )
--- a/ankihub/gui/browser/browser.py
+++ b/ankihub/gui/browser/browser.py
@@ -94,6 +94,7 @@ from .custom_search_nodes import (
     UpdatedInTheLastXDaysSearchNode,
     UpdatedSinceLastReviewSearchNode,
 )
+from .sidebar_tooltip import RICH_TOOLTIP_ATTR, setup_sidebar_rich_tooltip
 
 
 @dataclass
@@ -813,6 +814,7 @@ def _on_browser_did_search_handle_custom_search_parameters(ctx: SearchContext):
 # sidebar
 def _setup_ankihub_sidebar_tree():
     browser_will_build_tree.append(_on_browser_will_build_tree)
+    browser_will_show.append(lambda browser: setup_sidebar_rich_tooltip(browser.sidebar))
 
 
 def _on_browser_will_build_tree(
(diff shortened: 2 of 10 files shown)


cacoze (approved):
LGTM, can you send one addon version for QA before merging this one?

RisingOrange:
> LGTM, can you send one addon version for QA before merging this one?

Done, attached it here: https://ankihub.atlassian.net/browse/NRT-766?focusedCommentId=44053

PR #1300, opened 2026-05-27
[NRT-779] Suggest-a-change dialog QA fixes

RisingOrange:
## Related issues
- [NRT-779](https://ankihub.atlassian.net/browse/NRT-779) — Suggest-a-change dialog QA fixes.
- Follow-up to #1288 ([NRT-749](https://ankihub.atlassian.net/browse/NRT-749)) — addresses the QA pass on that PR.
- **Stacked on #1288** (base branch `NRT-749`); review the delta here.

## Proposed changes
QA-feedback fixes on the redesigned Suggest-a-change dialog, applied on top of #1288. Most issues only showed up on macOS, where the native checkbox renders differently than on Linux.

### Design polish
- The "Include in suggestion" panel and Source group now share a neutral background fill (light + dark) instead of a stroked border, matching Figma.
- Section dividers and source-input borders use a subtle theme-aware line.
- Uniform 16px gutter around the dialog (margins, between the two columns, between the content and the OK/Cancel row).
- **Source = "Other"** uses a 60px multi-line, top-aligned, line-wrapping input instead of a single-line field. Switching source types re-validates so the OK button gates correctly.
- The delete-note hint wraps to width instead of breaking mid-sentence.

### macOS-specific UI fixes
- Tag indicators no longer overlap their labels, and the indicator now lines up with the first line of the wrapped tag name.
- Tag rows occupy the same vertical box as field rows, so the visible gap between rows matches.
- Sub-checkboxes are no longer offset ~2px past the section title — section + items now share the same left edge.
- The section title's bold/muted styling no longer shifts its indicator relative to the items.
- The focus indicator that appeared on a checkbox label after clicking is suppressed on the panel's checkboxes (the focus visual on the indicator itself is kept; "Submit without review" stays as the Qt default).

### Behaviour fixes
- Clicking anywhere on a field/section/tag row now toggles it (the hover cursor previously implied clickable areas that didn't actually toggle).
- Ticking a section's select-all flips its children in a single visible update instead of a cascade of repaints.
- A section whose only row is the locked first field shows the select-all disabled and checked, so it can't toggle with no effect.

## How to reproduce
1. Enable the `auto_protect_fields_when_edited` feature flag.
2. Edit a field on an AnkiHub note → **Suggest a change**.
3. Confirm the panel has a neutral background (no border), thin grey section dividers, that clicking anywhere on a field/tag/section row toggles it, and that a long/hierarchical tag wraps with the label aligned to the first line of the indicator.
[...]

PR #1304, opened 2026-06-01
[NRT-751] Open the suggestion dialog in an empty state when there's nothing to suggest

RisingOrange:
## Related issues

- [NRT-751](https://ankihub.atlassian.net/browse/NRT-751) — sub-task of [NRT-508](https://ankihub.atlassian.net/browse/NRT-508), deferred from [NRT-749](https://ankihub.atlassian.net/browse/NRT-749).
- Builds on the NRT-749 suggestion-dialog redesign + QA fixes, which merged to `main` via #1300. This PR now targets `main` directly.
- Supersedes the earlier "disable the Suggest button" take (#1297) per the designer's decision on the ticket — keep the button enabled and handle the no-changes case in the dialog.

## Proposed changes

Per the updated NRT-508/751 spec: when a user opens the suggestion dialog for an AnkiHub note that has **nothing to suggest** (no field diff, no tag change), the dialog opens in an **empty state** instead of the button being disabled. The user can still pick **Delete** on the change-type dropdown to submit a deletion; the OK button stays disabled until there's a selection.

### Dialog (`gui/suggestion_dialog.py`)

- `IncludeInSuggestionWidget` renders an empty state when there are no edited fields and no tag changes: a warning icon + **"No changes detected"** + **"Edit a field to suggest a change, or select Delete on change type."**, the subtitle switches to **"Edit a note field to create a change suggestion"**, and the counter reads **0/0**. Everything else was already wired in NRT-749 — the single-suggestion path already opens the dialog unconditionally, OK is already gated on `has_any_selection()`, and DELETE already hides the field widget.
- The empty state renders in the panel **frame** (not the scroll area): a word-wrapped label inside the `widgetResizable` `QScrollArea` squashes/overlaps the subtitle at narrow widths. Hiding the scroll area (which only holds the checkbox list) and centering the empty state in the frame keeps it readable at any dialog size.
- **Pre-existing layout fix:** the "Source (Required)" group box shared the right column's vertical shrink, so its combo + link/details fields overlapped when the dialog was short. Pinned it to its content height (`SetMinimumSize` + `Fixed` vertical policy) so the rationale editor absorbs the flex instead.

### Editor (`gui/editor.py`)

- The **Suggest a change** button is enabled for any valid, non-deleted AnkiHub note; clicking opens the dialog, which handles the no-changes case via the empty state above.
- The keyboard-shortcut path fires even when the button is visually disabled, so it's gated for the cases it shouldn't act on:
  - non-AnkiHub note types → no-op (closes a pre-existing `AssertionError` from the dialog opener);
  - notes deleted on AnkiHub → tooltip *"This note has been deleted from AnkiHub. No new suggestions can be made."*;
  - new notes with an empty first field → tooltip *"The first field is empty."* (and skips `add_current_note`, so its add-note callback can't leak onto a later note).
- Enabled-state tooltip reworded from *"Send your request to AnkiHub"* to *"Suggest a change to AnkiHub"* (existing notes) / *"Suggest the note to AnkiHub"* (new notes).
- Button label/tooltip JS values are `json.dumps`-encoded so quotes/apostrophes can't break the script.

### Suggestions (`main/suggestions.py`)

- Promoted `_has_empty_first_field` to a public, shared `has_empty_first_field`, now used by the editor action gate, the suggestibility check (`_is_suggestible_from_diff`), and both submit paths (`suggest_note_update`, `_suggestions_for_notes`) — so "is this first field empty?" is decided one way everywhere.
- Strengthened that check: instead of a bare `.strip()`, it strips HTML tags + entities and treats embedded media as content, so markup-only values (`<br>`, `&nbsp;`) count as empty while an image/audio-only first field counts as filled. It's a pure-Python approximation of Anki's `fields_check` (reusing the shared `HTML_MEDIA_TAGS` / `AV_TAGS` regexes) to avoid a per-note backend round-trip; the server still rejects a truly-empty first field as a backstop.
- Added `globally_protected_fields_by_mid` to coerce the cached globally-protected fields into the typed shape the suggestibility helpers and the dialog widget consume.

## Screenshots

The empty state (no field diff, no tag change) — flat gray warning triangle, theme-adaptive copy:

<table><tr>
<td>Empty state on Linux (dark mode)</td>
<td>Empty state on macOS (dark mode)</td>
</tr><tr>
<td align="center"><em>Linux — dark</em></td>
<td align="center"><em>macOS — dark</em></td>
</tr></table>

## How to reproduce

1. Open an AnkiHub note with no local edits → click **Suggest a change** (now always enabled) → dialog opens with the **"No changes detected"** empty state, OK disabled.
2. Switch **Change Type → Delete** → the field panel hides and OK enables (with rationale filled) → deletion can be submitted.
3. Edit a field → the dialog shows the field checkbox list as before; OK enables once something's selected.
4. Resize the dialog short/narrow → the empty state stays centered and readable; the Source section no longer squashes.

## Tests

`TestEditor`: button stays enabled on an unchanged note; deleted-note action gate; non-AnkiHub no-op (logged in/out); new note with an empty first field doesn't register an add callback. `TestFieldsToSuggestFilters`: a widget empty-state test (no checkboxes, empty-state subtitle + warning, 0/0 counter, scroll area hidden, nothing selected). `TestSuggestNotesInBulk`: `has_empty_first_field` over empty/whitespace/markup-only values vs. text/image/audio content. All pass; mypy + ruff clean. Layout verified with a screenshot harness (since removed).

[NRT-751]: https://ankihub.atlassian.net/browse/NRT-751
[NRT-749]: https://ankihub.atlassian.net/browse/NRT-749
[NRT-508]: https://ankihub.atlassian.net/browse/NRT-508


cacoze:
@RisingOrange the font style for the texts `Change Type` and `Rationale for Change (Required)` seems a bit different from figma. They should have the same style of the text `Include in suggestion`, right? Maybe it's just the font weight that needs to be updated.
image

RisingOrange:
@cacoze 
> the font style for the texts Change Type and Rationale for Change (Required) seems a bit different from figma. They should have the same style of the text Include in suggestion, right? Maybe it's just the font weight that needs to be updated

Fixed! Yes, it was the font weight.

Now it looks like this:
image

cacoze:
Cool, the font weight issue was solved. Now I noticed this difference between this background on the source section. (Notice how the "gray" background is collapsing with the "Source (Required)" header:

Branch:
image

Figma:
image

RisingOrange:
@cacoze 
>  Now I noticed this difference between this background on the source section. (Notice how the "gray" background is collapsing with the "Source (Required)" header:

Oh yes, this is how titles are placed on `QGroupBox` elements in Qt. I don't think we need to change it, Maryna already saw this during a previous QA round and didn't have an issue with it

cacoze:
ok! LGTM than.

cacoze (approved):

PR #1305, opened 2026-06-01
[NRT-771] Post-flag-flip cleanup: drop AUTO_PROTECT_FEATURE_FLAG scaffolding + thread NoteDiff

RisingOrange:
## Related issues
- [NRT-771](https://ankihub.atlassian.net/browse/NRT-771)
- Follow-up to NRT-749 (#1288); origin write-up in `notes/nrt749-deferred-refactors.md` (#4).

## Proposed changes
`auto_protect_fields_when_edited` is now permanent for all users, so this removes the feature-flag scaffolding NRT-749 added. It also lands a related performance cleanup that was deliberately deferred until the flag was gone — reusing the per-note diff computed when the suggestion dialog opens instead of recomputing it on submit (see §3 for why the flag blocked it).

**1. Remove `AUTO_PROTECT_FEATURE_FLAG` scaffolding**
- Delete the constant in `main/suggestions.py`.
- Inline the 4 gate sites so the gated code always runs:
  - `editor.py` — auto-protect-on-unfocus hook (drops the early-return flag check).
  - `suggestion_dialog.py` — the bulk empty-state gate and the `_setup_ui` field-selection-widget gate.
  - `decks_dialog.py` — the "Automatically protect fields when edited" box is always shown.
- Both suggestion builders now pass `include_protected_fields=True` unconditionally.

**2. Tighten `BulkSuggestionFilters`**
- `fields_to_include_by_mid` is now required (drops `Optional`); `for_mid` no longer has a `None` "ship-everything" branch. The dialog's widget always renders and produces a concrete per-mid selection, so `None` was unreachable legacy. Empty-construction sites pass `{}`.

**3. One implementation of "what changed" instead of two (thread `NoteDiff` to the submit path)**

_Plain version:_ When you edit a note and open the suggestion dialog, the add-on computes a **`NoteDiff`** per note — *what changed* versus the version AnkiHub has on record (the changed fields, added/removed tags, new media). That diff is what powers the "Include in suggestion" picker. Previously it was thrown away once the dialog opened: when you hit submit, the builders re-derived the exact same information from scratch — re-converting each note, re-reading the AnkiHub DB, re-comparing fields and tags — plus an extra batch query. This change *threads* (passes along) that already-computed diff from the dialog through to the submit path, and the submit builders now just **apply your selection on top of the diff** instead of recomputing it: the diff is computed once and is the single source of truth, and a duplicate `AnkiHubNote.filter(...)` query is dropped.

_What this newly depends on:_ the diff is computed when the dialog opens and consumed when the user submits, so it must not go stale in between. The dialog is `ApplicationModal`, which is what makes that safe; nothing else enforces it, so keep that in mind if the dialog ever becomes non-modal.

_Why it was deferred:_ while the flag existed, the open-time conversion always included protected fields but the submit-time conversion only did so when the flag was on — so reusing the open-time diff could have shipped the wrong content. With the flag gone, both sides behave identically, so reuse is now safe. (Hence folding it into this ticket rather than NRT-749.)

_Details:_ `NoteDiff` carries `changed_fields: List[Field]`, the fields a suggestion would ship (changed-vs-AnkiHub for existing notes, non-empty for new notes), with `changed_field_names` as a derived property, plus `ah_nid` and the two membership booleans. It deliberately does *not* retain the source `NoteInfo` objects: an earlier revision held both the local note and the AnkiHub-stored one, which kept every field of every note alive for the dialog's lifetime at the 2000-note cap. Both the bulk and single-suggestion dialogs thread their open-time `compute_note_diffs` map to the submit path: `_suggestions_for_notes` classifies notes from the diff (`exists_in_ah_db` / `is_deleted_on_remote`, dropping the duplicate `AnkiHubNote.filter(anki_note_id__in=...)` batch query) and hands each note's diff to `_change_note_suggestion` / `_new_note_suggestion`. Those builders just apply the user's allowlist to `changed_fields` / `added_tags` / `removed_tags` — the field/tag comparison (`_fields_that_changed`, `_added_and_removed_tags`) lives only in `compute_note_diffs`. A builder self-computes the diff when one isn't supplied (standalone callers / tests). `BulkSuggestionFilters.fields_to_include_by_mid` is required, and within an explicit filter a missing per-mid entry ships none of that note type's fields (the dialog's per-section selection). Passing no filter at all (`filters=None`) means "no filter — ship everything the diff detected", matching the single-note path; the dialog always passes an explicit selection, so `None` is a programmatic convenience.

**4. Batch the media step's AnkiHub-DB reads**
Same submit path, one step further on from §3. `_rename_and_upload_media_for_suggestions` re-derived per note what the DB can answer per batch and per note type: `original_notes_data` called `ankihub_db.note_data` once per suggestion (two queries each), and `note_type_dict` (a query per call) was asked once per *note* by the media helpers though its answer only varies per *note type*. Now uses the batched `notes_data_for_anki_nids` (same not-deleted filter, so the media-name set is unchanged) and memoizes `note_type_dict` for the call. At the 2000-note bulk cap that is roughly 8000 queries removed, more than the diff threading itself removes. Worth calibrating, though: the bulk submit runs on a worker thread under a progress dialog, so this shortens a background step rather than unblocking the UI, and 2000 is the cap rather than a typical selection. At a realistic few-dozen-note suggestion the saving is milliseconds. The reason to do it is that the work was plainly redundant, not that it was slow.

**5. Name the confusable empty state**
`BulkSuggestionFilters.none_selected()` spells "ship nothing" at both construction sites, so it no longer reads as an unlabelled `{}` sitting next to a `filters=None` that means the opposite. The three-valued rule is stated once on `PerNoteFilters` instead of being restated at `for_mid`, `suggest_notes_in_bulk` and `_apply_field_allowlist`. `BulkSuggestionFilters`'s docstring now gives the real reason fields are keyed by note type: `last_deselected_fields_per_note_type` persists deselections at that granularity, so the widget's one-section-per-note-type layout follows the stored model rather than the other way round.

**Tests**
- Deleted the two flag-off legacy tests; dropped the `set_feature_flags({"auto_protect_fields_when_edited": ...})` enabling lines (the per-deck `set_auto_protect_fields_when_edited` setting stays).
- Dropped the now-dead flag-enabling line from a test added on `main` after this branch was cut.
- Pinned the bulk-DELETE behaviour change described under "How to reproduce" with a test, checked against `main`'s old semantics, where it fails.
- Added two regression guards for §3/§4, which are otherwise unpinned: one asserting `note_data` and `notes_data_for_anki_nids` agree about deleted rows (the equivalence the batched read rides on), and one asserting the AnkiHub-DB read count for a bulk submit does not grow with the number of notes. The second counts queries at the peewee layer rather than spying on particular reader methods, so a per-note read reintroduced through any helper is caught; it was checked against a per-note regression, where it fails, so it is not a vacuous guard.
- Gave bulk-path tests explicit `fields_to_include_by_mid` (the old `None` "ship-everything" default now ships nothing), and updated the shared `mock_suggestion_dialog` fixture to build its allowlist from the diffs — mirroring the real widget's default-all-checked selection.

## How to reproduce
Almost entirely an internal refactor — given the flag is on for everyone, there is no user-facing change on the ordinary suggest paths. Covered by tests.

One deliberate exception, worth a reviewer's eye. Because the field-selection widget is hidden for DELETE, `suggestion_meta()` falls back to an empty selection, and the empty per-mid map now means "ship no fields" where the old `None` meant "ship everything". So a **bulk DELETE over a selection that includes notes not yet on AnkiHub** now reports those notes as "no changes", where before it submitted full *new-note* suggestions for them (browser multi-select including a non-AnkiHub note → Bulk suggest → change type "Delete note" → submit). Submitting a new-note suggestion in response to a delete request was never intended; this is the sane behavior, but it is a change.

## Further comments
Deliberately left `PerNoteFilters` / `suggest_note_update` / `suggest_new_note` `filters=None` "ship-everything" default in place: unlike the `BulkSuggestionFilters` map (only ever populated by the dialog widget), that's a sensible default for the general-purpose single-note suggest entry points, not flag residue.

[NRT-771]: https://ankihub.atlassian.net/browse/NRT-771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ankihub/main/suggestions.py
```diff
@@ -140,55 +136,54 @@ class PerNoteFilters:
...
 
-    fields_to_include_by_mid: Optional[Mapping[NotetypeId, Collection[str]]] = None
+    fields_to_include_by_mid: Mapping[NotetypeId, Collection[str]]
     tags_to_add: Optional[Collection[str]] = None
     tags_to_remove: Optional[Collection[str]] = None
 
+    @classmethod
+    def none_selected(cls) -> "BulkSuggestionFilters":
+        """Distinct from passing no filters at all, which ships everything."""
+        return cls(fields_to_include_by_mid={})
```
cacoze: `none_selected()` only empties `fields_to_include_by_mid`. `tags_to_add` / `tags_to_remove` stay `None` ("ship all tags"), so the name oversells "nothing selected."

Harmless on the DELETE fallback — `_change_note_suggestion` ignores tags, and new notes die on the first-field check — but empty tuples here would match the name and the three-valued rule on `PerNoteFilters`.

ankihub/gui/suggestion_dialog.py
```diff
@@ -114,7 +113,14 @@ class SuggestionMetadata:
...
     source: Optional[SuggestionSource] = None
-    filters: BulkSuggestionFilters = field(default_factory=BulkSuggestionFilters)
+    filters: BulkSuggestionFilters = field(default_factory=BulkSuggestionFilters.none_selected)
+
+
+def _has_live_ah_note(diff: NoteDiff) -> bool:
+    """Whether AnkiHub has a note for this one that can still take a change suggestion.
+    A note deleted on AnkiHub is a new-note candidate again, not a change candidate.
+    """
+    return diff.exists_in_ah_db and not diff.is_deleted_on_remote
```
cacoze, line 125: Nice helper — this is the right predicate (matches the old `ankihub_nid_for_anki_nid` / `NOTE_NOT_DELETED` check).

The first-field lock in `IncludeInSuggestionWidget._populate` still uses `not diff.exists_in_ah_db`, so a deleted-on-remote note in a mixed batch wouldn't lock the first field. Pre-existing, and those notes aren't submitted in bulk anyway. Easy follow-up: use `_has_live_ah_note` there too.

ankihub/gui/suggestion_dialog.py
```diff
@@ -114,7 +113,14 @@ class SuggestionMetadata:
     auto_accept: bool = False
     change_type: Optional[SuggestionType] = None
     source: Optional[SuggestionSource] = None
-    filters: BulkSuggestionFilters = field(default_factory=BulkSuggestionFilters)
+    filters: BulkSuggestionFilters = field(default_factory=BulkSuggestionFilters.none_selected)
```
cacoze, line 118: Default is now "ship nothing" (`none_selected()`), which is the right DELETE fallback but an easy footgun if someone constructs `SuggestionMetadata` without `filters`.

The dialog and `mock_suggestion_dialog` were updated. `TestSuggestionDialog.test_visibility_of_form_elements_and_form_result` omits `filters=` and only passes because both sides share this default — worth a one-line note on the field, or making `filters` required so a missing allowlist fails loud.

ankihub/main/suggestions.py
```diff
@@ -325,22 +322,22 @@ def suggest_note_update(
...
 ) -> ChangeSuggestionResult:
     """Sends a ChangeNoteSuggestion to AnkiHub if the passed note has changes.
     Also renames media files in the Anki collection and the media folder and uploads them to AnkiHub.
     Returns a ChangeSuggestionResult enum value.
 
-    `filters` carries the user's optional allowlists for fields and added/removed
-    tags. `None` for any list means "no filter for that dimension"; the absent
-    `filters` arg is equivalent to no filters at all.
+    `diff` is the note's precomputed `NoteDiff`; the dialog passes its open-time copy,
+    which is why the note must not change between opening the dialog and submitting.
```
cacoze, line 332: Small doc regression: the `filters` explanation was dropped when `diff` was added. The three-valued `None`-vs-empty rule is easy to get wrong (see `PerNoteFilters`), so I'd keep a line on `filters` here too.

ankihub/main/suggestions.py
```diff
@@ -864,16 +843,22 @@ def _rename_and_upload_media_for_suggestions(
...
+    anki_nids = [NoteId(s.anki_nid) for s in suggestions]
+
+    note_type_dicts: Dict[NotetypeId, NotetypeDict] = {}
+
+    def note_type_dict_for_mid(mid: int) -> NotetypeDict:
+        # The media helpers ask per note, but the answer only varies per note type.
+        key = NotetypeId(mid)
+        if key not in note_type_dicts:
+            note_type_dicts[key] = ankihub_db.note_type_dict(key)
+        return note_type_dicts[key]
```
cacoze, line 855: Nit: this memo is per `_rename_and_upload_media_for_suggestions` call, so the new-note and change-note lists don't share it — one extra `note_type_dict` query per list, not per note. Totally fine given the ~8000 queries this removes at the cap; only mentioning it because a function-level cache (or passing the dict in) would close the leftover if you ever touch this again.

cacoze (commented):
Non-blocking nits from review — none of these should block merge. Inline notes below.

cacoze (dismissed):
I`ve Tested on my end and LGTM. The comments I`ve left are not blocker, but I'm happy to review/approve again if you want to address any of those

cacoze (approved):

PR #1309, opened 2026-06-08
[NRT-790] Fix long-tag clipping and make tag/column display responsive in the suggestion dialog

RisingOrange:
## Related issues
- [NRT-790](https://ankihub.atlassian.net/browse/NRT-790)

### 1. Fix clipped last line of long tag checkboxes

In the note suggestion dialog ("Include in suggestion" panel, behind `auto_protect_fields_when_edited`), long tag checkboxes had their last line clipped:

- A deep hierarchical tag like `#AK_MCAT_v2::MileDown::Behavioral::Cognition` lost its bottom segment ("Cognition") — the row wrapped to one more line than its reserved height. The clip appeared/disappeared at boundary widths when resizing the dialog.
- An elided tag whose leaf has no break points overflowed horizontally off the right edge.

More noticeable with larger UI fonts.

**Root cause:** `_WrappingCheckBox.heightForWidth` measured text wrapping against a wider width than `paintEvent` actually drew into. The style content-rect metrics were cached in `__init__`, *before the widget was polished* — and the dialog's stylesheet activates `QStyleSheetStyle`, which pads the checkbox content rect once polished, so the cached width was ~6px too wide and `heightForWidth` reserved one line too few. Wrapping was also measured with `QFontMetrics.boundingRect`, a different code path than the painter, which disagrees at wrap boundaries. This is why it only reproduced in the real dialog, not in an isolated widget.

**Fix:**
- Compute the label content width fresh per call (the same `SE_CheckBoxContents` the painter uses) instead of caching a pre-polish value.
- Share one `QTextLayout` between `heightForWidth` and `paintEvent` so the reserved height matches the painted line count exactly.
- Wrap at word boundary **or anywhere** so an over-long unbreakable leaf wraps instead of overflowing horizontally.

### 2. Scale both columns proportionally when resizing (review feedback)

Previously, enlarging the dialog gave almost all extra width to the right (Change Type) column: the "Include in suggestion" panel was capped at 360px, even though it's the column that benefits most from extra room (long tags). The columns already had 2:3 stretch factors — the cap just kept them from applying. Dropping the cap lets both columns grow proportionally (~40/60). Long content still can't push the dialog wide: section titles and tag rows elide (`Ignored` size policy) and the rest sits inside a scroll area.

### 3. Reveal more of an elided tag as the dialog widens (QA feedback)

A long hierarchical tag was always collapsed to `…::leaf` (an 85-char display ceiling from the [NRT-508](https://ankihub.atlassian.net/browse/NRT-508) spec), regardless of how much room the panel had — so even at a very wide window the leading segments stayed hidden. The bound is now **space, not a character count**: `_WrappingCheckBox` renders the longest candidate (full tag → progressively shorter `…::`-suffixes) that fits within a **3-line budget** at the current width. So widening the dialog progressively reveals more (`…::VariableRatio` → `…::ReinforcementSchedules::VariableRatio` → the **full tag** once it fits three lines), and narrowing collapses back toward `…::leaf`. A single segment too long to fit the budget is truncated with a trailing `…` so a pathological tag can't balloon the row. The full tag remains available as a tooltip while elided.

## How to reproduce
1. Enable the `auto_protect_fields_when_edited` feature flag.
2. On a note in an AnkiHub deck, remove a deep tag such as `#AK_MCAT_v2::MileDown::Behavioral::Cognition` (and/or add a tag with a very long final segment).
3. Open the suggestion dialog and look at the "Removed Tags" / "Added Tags" checkboxes; resize the dialog.
4. Before this fix, the last tag segment is cut off (and reappears only at certain widths). After, every wrapped line stays fully visible.
5. Enlarge the dialog window: before, the left panel stopped growing at 360px; after, it keeps ~40% of the width and long tags become fully readable.
6. Add a long multi-level tag (e.g. `#AK_MCAT_v2::Psychology::LearningAndMemory::OperantConditioning::ReinforcementSchedules::VariableRatio`) and widen the dialog: the `…::` portion reveals more segments as space grows, and a wide-enough window shows the whole tag with no `…::` at all.

## Screenshots and videos
Suggestion dialog with a deep tag and a long-leaf tag, before and after the clipping fix:

<table><tr>
<td><strong>Before</strong><br/>Before: last tag segment clipped</td>
<td><strong>After</strong><br/>After: tags wrap fully</td>
</tr></table>

Proportional column scaling at a wide window size (1500px):

<table><tr>
<td><strong>Before: left panel capped at 360px</strong><br/>Before: left panel capped at 360px at wide window sizes</td>
</tr><tr>
<td><strong>After: columns scale proportionally</strong><br/>After: both columns scale proportionally</td>
</tr></table>

Long tags revealing more as the dialog widens — narrow: tags wrap (up to 3 lines) and elide with `…::`; wide: the same tags shown in full:

<table><tr>
<td valign="top"><strong>Narrow (elided)</strong><br/>Narrow window: long tags wrap to up to 3 lines and elide with a leading ellipsis</td>
<td valign="top"><strong>Wide (full)</strong><br/>Wide window: the same tags shown in full</td>
</tr></table>

## Further comments
- Tests: added `TestFieldsToSuggestFilters::test_long_tag_checkbox_is_not_clipped`, a dialog-level regression test that reproduces both the vertical (deep tag) and horizontal (long leaf) clipping in the real dialog and asserts no clipping across a range of widths. Added `test_tag_label_elision_candidates`, `test_wrapping_checkbox_reveals_more_segments_when_wider` (elides when narrow, shows the full tag when wide), and `test_wrapping_checkbox_truncates_giant_single_segment` for the width-responsive elision. Verified the clipping test fails on the buggy version and passes on the fix. `mypy` clean.
- The per-call style query + `QTextLayout` build replace a cache that was the source of the bug; reviewers (Codex + cleanup pass) confirmed the cost is negligible at realistic tag counts. The width-responsive elision adds a small per-resize candidate search (≤ number of `::` segments, memoized per width).
- Verified on both Linux and macOS (native style differs there, and the fix depends on the native checkbox content-rect metrics) — tag checkboxes render without clipping across dialog widths.
- Changes 2 and 3 came from review/QA feedback on this PR; the suggestion/dialog integration tests and the wrapping/clipping unit tests pass with them.

[NRT-790]: https://ankihub.atlassian.net/browse/NRT-790?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[NRT-508]: https://ankihub.atlassian.net/browse/NRT-508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddevdan (commented):
LGTM  but we need to create the addon for Maryna/Daemon to do QA

ddevdan (dismissed):
LGTM
But, just a heads up 
Next steps before merging:
1. Create PR env 
2. Set the feature flag there
2. Create addon pointing to the PR env
3. Move to Ready for QA

ddevdan:
I was wondering if we could fix this, but I think it's more of a bug on Anki's side than in the add-on.

image

RisingOrange:
> I was wondering if we could fix this, but I think it's more of a bug on Anki's side than in the add-on.

Yeah, that's an issue with Anki. I don't think we need to do anything about it atm

RisingOrange:
@ddevdan Maryna asked for another improvement ("Reveal more of an elided tag as the dialog widens") , so I added it to this PR

ddevdan (approved):

PR #1313, opened 2026-06-16
Fix Unicode line separators in fields causing JSON errors

abdnh:
## Related issues

https://community.ankihub.net/t/anatoking-download-error/601183/

https://ankihub.sentry.io/issues/7447487725/?project=6546414&query=id%3A33321a0acae34a17a2f85709d3f9185a&referrer=issue-stream

## Proposed changes

str.splitlines() is used to split the CSV content into lines, but this function [also splits on unprintable characters such as Line Separator](https://docs.python.org/3/library/stdtypes.html#str.splitlines
), which causes errors if some field happens to contain these characters.

### Before

Import the [AnatoKing](https://app.ankihub.net/decks/bea25743-706a-4bff-a043-efcc0a617121) deck and confirms you get a JSON parsing error.

### After

Confirm deck import works with no errors.

### DIFF
--- a/ankihub/ankihub_client/ankihub_client.py
+++ b/ankihub/ankihub_client/ankihub_client.py
@@ -682,7 +682,7 @@ class AnkiHubClient:
         else:
             deck_csv_content = s3_response_content.decode("utf-8")
 
-        reader = csv.DictReader(deck_csv_content.splitlines(), delimiter=CSV_DELIMITER, quotechar="'")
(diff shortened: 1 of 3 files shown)


abdnh:
Looks like the deck also has newlines characters in the fields, so this fix is not sufficient. We have to strip all characters recognized by .splitlines() on the website side. Will create a task for that. I fixed the deck manually for now.

PR #1314, opened 2026-06-17
[NRT-799] Improve suggestion dialog copy and protected-field discoverability

RisingOrange:
## Related issues
- [x] Closes [NRT-799](https://ankihub.atlassian.net/browse/NRT-799)

## Proposed changes
Addresses Ahmed's feedback on the protected-fields selection experience (NRT-799).

**1. Suggestion dialog copy** — the "Include in suggestion" wording was unclear without reading the supporting text:
- Title: `Include in suggestion` → **`Select fields to include`**
- Subtitle: `Select which fields you want to submit changes for` → **`Choose which fields to submit changes for.`**

**2. Protected-field discoverability** — in the browser's "Protect fields" dialog, globally-protected fields render checked + greyed out with an explanatory tooltip, but there was no visual cue the tooltip existed (it only appeared on hover). Added an **info icon after the field name** for those locked rows so the tooltip is discoverable.

Implementation note: Anki's Qt style lays out the checkbox + text inside `drawControl`, so the `QStyleOptionViewItem` handed to a delegate's `paint()` has no usable sub-element geometry. The new `_InfoIconDelegate` re-runs `initStyleOption` on a copy of the option before calling `subElementRect(SE_ItemViewItemText)`, which is what lets it place the icon immediately after the rendered text rather than at the row's far-right edge.

**3. Test-stability fix (drive-by)** — while iterating I hit a pre-existing flaky teardown: the shared `anki_session_with_addon_data` fixture points `ANKIHUB_BASE_PATH` at a `TemporaryDirectory`, and background tasks (e.g. `refresh_user_state_in_background`, exercised by `TestFeatureFlags`) can still be writing config/logs into it when a test ends — so the strict `rmtree` races them and intermittently fails at teardown with `OSError: [Errno 39] Directory not empty`. These disposable isolation dirs are now cleaned up with `shutil.rmtree(ignore_errors=True)` in `conftest.py` and the media-folder fixture (the other site exposed to background writers). The test bodies are unaffected — only the cleanup is made race-tolerant.

## How to reproduce
**Suggestion dialog copy:**
1. Open a deck synced with AnkiHub, edit a field on a note.
2. Right-click → AnkiHub → Suggest a change (or the editor button).
3. The "Select fields to include" panel shows the new title + subtitle.

**Protected-field icon:**
1. Have an AnkiHub deck with at least one globally-protected field (set on the AnkiHub website).
2. In the browser, select a note of that note type → right-click → AnkiHub → Protect fields.
3. The globally-protected field rows are greyed/checked and now show an info icon right after the field name; hovering reveals the "This field is globally protected" tooltip.

## Screenshots and videos

**Suggestion dialog — new title + subtitle**

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>Suggestion dialog before</td>
<td>Suggestion dialog after</td>
</tr>
</table>

**Protect fields dialog — info icon after globally-protected field names**

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>Protect fields dialog before</td>
<td>Protect fields dialog after</td>
</tr>
</table>

## Further comments
The info icon is positioned with a small `_InfoIconDelegate` rather than a leading item icon, because we want it *after* the field name. Getting "after the text" right cross-platform needs two style-geometry details: the option must be re-`initStyleOption`'d before `subElementRect(...ItemText)` returns usable geometry, and the text's left inset (`PM_FocusFrameHMargin`) must be added — it's larger on macOS than on Fusion, and omitting it made the icon crowd the text there. Long field names elide before the icon (space is reserved before the native paint).

Verified on **Linux** (headless render + `protect_fields_action` integration tests pass, mypy clean) and on **macOS** (built + installed on a Mac; icon spacing confirmed after the `PM_FocusFrameHMargin` fix).

[NRT-799]: https://ankihub.atlassian.net/browse/NRT-799?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/suggestion_dialog.py
+++ b/ankihub/gui/suggestion_dialog.py
@@ -84,9 +84,10 @@ from .utils import (
     warning_triangle_icon,
 )
 
-# "Include in suggestion" panel copy. The subtitle switches to the EMPTY variant
+# "Select fields to include" panel copy. The subtitle switches to the EMPTY variant
 # (plus the warning) when the note has no edited fields and no tag changes.
-INCLUDE_SUBTITLE = "Select which fields you want to submit changes for"
+INCLUDE_TITLE = "Select fields to include"
+INCLUDE_SUBTITLE = "Choose which fields to submit changes for."
 EMPTY_STATE_SUBTITLE = "Edit a note field to create a change suggestion"
 EMPTY_STATE_TITLE = "No changes detected"
 EMPTY_STATE_HINT = "Edit a field to suggest a change, or select Delete on change type."
@@ -1429,7 +1430,7 @@ class IncludeInSuggestionWidget(QWidget):
         frame.setLayout(frame_layout)
 
         title_row = QHBoxLayout()
-        self._title_label = QLabel("<b>Include in suggestion</b>")
+        self._title_label = QLabel(f"<b>{INCLUDE_TITLE}</b>")
         self._counter_label = QLabel("")
         counter_font = self._counter_label.font()
         counter_font.setPointSize(max(counter_font.pointSize() - 2, 8))
--- a/ankihub/gui/suggestion_dialog.py
+++ b/ankihub/gui/suggestion_dialog.py
@@ -487,7 +487,7 @@ class SuggestionDialog(QDialog):
         super().__init__(parent)
         # Block input to other Anki windows while the dialog is open — prevents
         # the user from editing the selected notes (or syncing) between dialog
-        # open and submit, which would invalidate the "Include in suggestion"
+        # open and submit, which would invalidate the "Select fields to include"
         # selection.
         self.setWindowModality(Qt.WindowModality.ApplicationModal)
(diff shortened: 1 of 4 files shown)


cacoze:
Nit (outside this diff): `tests/addon/test_integration.py` line 2244 still refers to an empty **"Include in suggestion"** panel — could update to **"Select fields to include"** to match the copy change in this PR.

cacoze (dismissed):
I've left a comment about updating a line that is refering to an outdated copy. It doesn`t need to block the merge.

RisingOrange:
Good catch, fixed in bcc509ed. Also updated the matching comment in `suggestion_dialog.py` that referenced the old panel name.

cacoze (approved):

PR #1325, opened 2026-07-14
[NRT-836] QA1 Smart Search

ddevdan:
## Related issues
- Closes [NRT-836](https://ankihub.atlassian.net/browse/NRT-836)

## Proposed changes
- Change title of the window to "Ankihub | Smart Search"
- Fix the flashcard selector appearing to "close" when the file picker is dismissed. It never actually closed: on macOS, dismissing the native file picker opened from the web view (whether cancelling or selecting a file) activates Anki's main window instead of the non-modal dialog the picker was opened from, burying the dialog behind it.
- Inject a JS watcher on page load in `AnkiHubWebViewDialog` that reports when a file input's picker closes (`cancel`/`change` events), and raise + activate the dialog in response. Lives in the shared base dialog, so it covers any AnkiHub web-view dialog that opens a file picker.

## How to reproduce
1. Open the flashcard selector for a deck with note embeddings.
2. Click the file picker (attach file) button — the native OS file picker opens.
3. Click **Cancel** (or select a file).
4. Before the fix: the selector appears to close (it's actually hidden behind the main Anki window). After the fix: the selector stays in front.

## Further comments
Root-caused via instrumented repro runs: no code path (ours, Anki's, or the web app's) ever closed the dialog — every close/hide/Escape/pycmd path was ruled out, and a window-state probe showed the dialog `visible=True, exposed=False` (occluded) while the main window became active. The regression test (`test_dialog_is_raised_when_file_picker_closes`) exercises the real seam (watcher JS → `cancel` event → pycmd → dialog raised); note it runs in CI on Linux, as the `pytest-forked` webview harness segfaults on macOS. The underlying macOS focus behavior itself is only observable manually — worth a one-time QA check on macOS.

## Screenshots

[NRT-836]: https://ankihub.atlassian.net/browse/NRT-836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/flashcard_selector_dialog.py
+++ b/ankihub/gui/flashcard_selector_dialog.py
@@ -9,6 +9,7 @@ from aqt.utils import openLink
 from .. import LOGGER
 from ..gui.webview import AnkiHubWebViewDialog
 from ..settings import (
+    config,
     url_flashcard_selector,
     url_flashcard_selector_embed,
     url_plans_page,
@@ -43,7 +44,10 @@ class FlashCardSelectorDialog(AnkiHubWebViewDialog):
         return cls.dialog
 
     def _setup_ui(self) -> None:
-        self.setWindowTitle("AnkiHub | Flashcard Selector")
+        if config.get_feature_flags().get("smart_search_revamp", False):
+            self.setWindowTitle("AnkiHub | Smart Search")
+        else:
+            self.setWindowTitle("AnkiHub | Flashcard Selector")
         self.resize(1000, 800)
 
         super()._setup_ui()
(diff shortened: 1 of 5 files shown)


cacoze (dismissed):

cacoze:
Nit: `test_dialog_is_raised_when_browser_opened_from_it_closes` has `@pytest.mark.sequential` duplicated (two stacked instances) — likely a rebase/cherry-pick artifact. Harmless functionally, but worth deduplicating:

https://github.com/AnkiHubSoftware/ankihub_addon/blob/2132e60/tests/addon/test_integration.py#L8765-L8766

cacoze (approved):

PR #1329, opened 2026-07-22
[KNW-174] Add a media sync progress indicator to the main screen

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/KNW-174

## Proposed changes

This adds a media sync indicator to the top toolbar, shown when a media sync is active.

## How to reproduce

- Trigger an AnkiHub sync from AnkiHub>Sync with AnkiHub.
- Once the blocking progress dialog finishes, the top toolbar should update to show the background media sync indicator.
- Hover over the toolbar button and confirm a short status label is shown.
- Confirm it matches with the AnkiHub>Media sync indicator (this is not modified for now, but we may want to make it interactive later).

## Screenshots and videos

image
image
image
image

## Further comments

Tests will be added in KNW-175


cjasm (approved):
LGTM

PR #1339, opened 2026-07-27
[KNW-175] Implement a dialog to display media sync progress

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/KNW-175

## Proposed changes

Implements a dialog to show media sync progress and any errors.

## How to reproduce

1. Trigger a media sync in any way (e.g. install a deck, upload a new deck, suggest changes with new media)
2. Click on the button at the top toolbar added in #1329
3. Inspect media progress and test the different buttons. To test error states, you can add dummy exceptions to `_download_media` and `_zip_and_upload_media_chunk`.

## Screenshots and videos
image
image
image
image


ankihub/gui/media_sync.py
```diff
@@ -308,8 +380,156 @@ def _set_toolbar_button_status(self, status: MediaSyncStatus) -> None:
...
+            self.toggle_log_button.setText("Hide log")
+            self.error_log_area.show()
+            self.error_log_spacer.hide()
+        else:
+            self.toggle_log_button.setText("Show log")
+            self.error_log_area.hide()
+            self.error_log_spacer.show()
+
+    def _on_retry(self) -> None:
+        assert media_sync._last_op_callback is not None
```
fernandolins, line 473: This assertion doesn't make sense here. What did you mean?

fernandolins (approved):
LGTM - with just one caveat regarding the assertion

cjasm (approved):

ankihub/gui/media_sync.py
abdnh: It's normally required for mypy, but looks like we have `strict_optional = false` set, so this does nothing now.

PR #1341, opened 2026-07-28
[KNW-197, KNW-208, KNW-209]Center error icon and message in AnkiWeb dialog, sizing to text length

fernandolins:
## Related issues
- [x] Closes [KNW-197](https://ankihub.atlassian.net/browse/KNW-197)
- [x] Closes [KNW-208](https://ankihub.atlassian.net/browse/KNW-208)
- [x] Closes [KNW-208](https://ankihub.atlassian.net/browse/KNW-209)

## Proposed changes
Fixes the error message row in the AnkiWeb sign-in/sign-up dialog (`ErrorLabel` in `ankihub/gui/ankiweb.py`), where the error icon and the error text were not visually centered as a single block.

Now:
- `status` gets a shrink-to-fit width, computed in `_update_status_width()` by measuring the label's own `sizeHint()` with word wrap temporarily disabled (this matches exactly how Qt renders the rich-text label on a single line, which a manual `QFontMetrics` calculation didn't reliably reproduce).
- That width is capped by `_max_status_width()`, based on the dialog's known fixed content width (the dialog uses `FixedDialogLayout` with a fixed 525px width), since querying `parentWidget().width()` at the time `set_error()` runs isn't reliable (the layout may not have been activated yet).
- `hbox.setAlignment(AlignCenter)` now correctly centers the icon+text row as a whole, since the row has a deterministic size instead of an expanding one.

As a result, short error messages appear as a compact, centered block next to the icon, while longer messages still wrap and use the full available width.

## How to reproduce
1. Open the "Sign into AnkiWeb" dialog (e.g. via the AnkiHub menu or sync flow).
2. Trigger any inline error in the form (e.g. request a login code, then enter an invalid/expired code and submit, or enter more requests than the throttle limit allows).
3. Before this change, the error icon appeared pinned to the left while the error text was centered only in the space to its right, making the pair look off-center. After this change, the icon and text are centered together as a single block, sized to the text.

## Further comments
This is a small, self-contained UI fix in `ErrorLabel._setup_ui`/`set_error`. No other files are affected by this change.

[KNW-197]: https://ankihub.atlassian.net/browse/KNW-197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KNW-208]: https://ankihub.atlassian.net/browse/KNW-208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[KNW-208]: https://ankihub.atlassian.net/browse/KNW-208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/ankiweb.py
+++ b/ankihub/gui/ankiweb.py
@@ -154,6 +154,10 @@ class LabelWithLink(QLabel):
 
 
 class ErrorLabel(QWidget):
+    _ICON_SIZE = 20
+    # FormWidget's inner content width, given the dialog's fixed 525px width and margins.
+    _FALLBACK_MAX_WIDTH = 461
+
     def __init__(self, parent: QWidget | None = None):
         super().__init__(parent)
         self._setup_ui()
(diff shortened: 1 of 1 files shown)


Luc-Mcgrady (approved):
For others reference here is the difference.

Old/With this PR
Image

abdnh (approved):

ankihub/gui/ankiweb.py
```diff
@@ -163,22 +167,48 @@ def _setup_ui(self) -> None:
...
+        # a manual QFontMetrics calculation doesn't reliably reproduce (e.g. it
+        # ignores HTML formatting and any stylesheet font overrides applied at
+        # render time).
+        status = self.status
+        status.setWordWrap(False)
+        natural_width = status.sizeHint().width()
+        status.setWordWrap(True)
+        # Small buffer to avoid the rare case where sizeHint() is one pixel short
+        # of what's needed, which would otherwise force an unwanted wrap.
+        self.status.setFixedWidth(min(natural_width + 2, self._max_status_width()))
```
abdnh, line 204: While testing #1355, I noticed that the buffer continuously pushes the icon to the left with each call. Not a big issue, just writing it down in case someone has a solution.

PR #1343, opened 2026-07-29
[KNW-207] Add back button

Luc-Mcgrady:
https://ankihub.atlassian.net/browse/KNW-207

image
image

Enter an email that already exists or enter an email ready to sign up and you should see a back button that will bring you back to the signup page.


fernandolins:
It looks good, but I found an inconsistency.
If you try to create a new account using the password.
But it uses an email for which an account already exists. You will be taken to the existing account screen. However, when you click the back button, the button redirects to the sign-up screen with a magic code.

Should we wait for this, or should we go back to the sign-up screen with a password?

abdnh (approved):

fernandolins (approved):
LGTM

PR #1330, opened 2026-07-22
[KNW-188] Add AnkiWeb client for login/signup

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/KNW-188
https://ankihub.atlassian.net/browse/KNW-189

## Proposed changes

Extend AnkiHubClient with AnkiWeb support and hook the methods into the the new login/signup screen implemented in #1324

## How to reproduce

- Run `scripts/build.py` to generate required Protobuf models.
- Run AnkiWeb locally and launch the add-on with the `ANKIWEB_URL` env var set.
- Test the different login/signup flows and error states.
- (Optional) to test syncing, run something like this in the [debug console](https://docs.ankiweb.net/misc.html#debug-console) to tell Anki to use a custom sync server:
```python
import aqt
aqt.mw.pm.set_custom_sync_url("https://ankiweb.test")
```

PR #1349, opened 2026-07-30
Set custom sync url to ANKIWEB_URL

Luc-Mcgrady:
Calls `aqt.mw.pm.set_custom_sync_url(override_url)` where `override_url` is the `ANKIWEB_URL` environment variable.

See https://github.com/AnkiHubSoftware/ankihub_addon/pull/1330

### DIFF
--- a/ankihub/entry_point.py
+++ b/ankihub/entry_point.py
@@ -11,9 +11,8 @@ from anki.hooks import wrap
 from aqt.gui_hooks import profile_did_open, profile_will_close, sync_did_finish
 from aqt.main import AnkiQt
 
-from .ankihub_client import DEFAULT_ANKIWEB_URL
-
 from . import LOGGER, anki_logger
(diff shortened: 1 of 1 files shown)


ankihub/entry_point.py
```diff
@@ -11,6 +11,8 @@
 from aqt.gui_hooks import profile_did_open, profile_will_close, sync_did_finish
 from aqt.main import AnkiQt
 
+from ankihub.ankihub_client.ankihub_client import DEFAULT_ANKIWEB_URL
```
iamllama:
```suggestion
from .ankihub_client import DEFAULT_ANKIWEB_URL
```
Fixes the CI lint check

iamllama: Ruff's still complaining about the import order 😅, try `just lint`

abdnh (approved):

PR #1350, opened 2026-07-30
[KNW-176] Improve reporting of media sync cancellation

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/KNW-176

## Proposed changes

This improves reporting of media sync cancellation:
- Add `MediaSyncStatus.CANCELING` for consistent tracking of cancellation.
- Reset sync state in the download/upload task's finish handler, rather than when cancellation is requested. Otherwise a subsequent sync can fight over the progress bar or report "idle" (because we're resetting _download_in_progress and _amount_uploads_in_progress).
- Add more cancellation checks so that cancel requests are responded to faster.

The original goal of KNW-176 was to investigate a rare issue where the media sync completes without downloading all missing files. I failed to find any clues about it.

## How to reproduce

- Install a deck with a lot of media files.
- Click the sync button on the toolbar then cancel the sync.
- Confirm the status is updated to "Canceling..." then "Finished".


ankihub/gui/media_sync.py
```diff
@@ -336,10 +341,12 @@ def _get_status(self) -> MediaSyncStatus:
             status = MediaSyncStatus.IDLE
         return status
 
-    def _refresh_media_download_status_inner(self):
+    def _refresh_media_download_status_inner(self, is_retry: bool):
         status = self._get_status()
+        print("_refresh_media_download_status_inner", status)
```
fernandolins: Could you remove it please?

ankihub/ankihub_client/ankihub_client.py
```diff
@@ -521,6 +521,8 @@ def upload_media(
         with ThreadPoolExecutor(max_workers=THREAD_POOL_MAX_WORKERS) as executor:
             futures: List[Future] = []
             for chunk_number, chunk in enumerate(media_path_chunks):
+                if self.should_stop_background_threads:
+                    LOGGER.info("Background threads stopped, aborting upload tasks...")
```
fernandolins, line 525: Should this `If` condition have a break or return?

abdnh: Yes!

fernandolins (dismissed):
LGTM - I'm only missing a test for this. It would be great to guarantee how sync should work.

ankihub/gui/media_sync.py
```diff
@@ -520,8 +528,17 @@ def increment_progress(self, increment: int = 1) -> None:
...
+
     def update_count_label(self) -> None:
-        self.count_label.setText(f"{self.progress_bar.value()}/{self.progress_bar.maximum()} files")
+        if self.progress_bar.maximum():
+            label = f"{self.progress_bar.value()}/{self.progress_bar.maximum()} files"
+        else:
+            label = ""
+        self.count_label.setText(label)
 
     def set_maximum(self, maximum: int) -> None:
```
fernandolins: None is calling it. Is it right? if so, can you remove it?

ankihub/gui/media_sync.py
```diff
@@ -336,10 +341,11 @@ def _get_status(self) -> MediaSyncStatus:
             status = MediaSyncStatus.IDLE
         return status
 
-    def _refresh_media_download_status_inner(self):
+    def _refresh_media_download_status_inner(self, is_retry: bool):
         status = self._get_status()
         self._set_status_text(status)
         self._set_toolbar_button_status(status)
+        self._dialog.update_status(self._get_status(), is_retry=is_retry)
```
fernandolins:
```suggestion
        self._dialog.update_status(status, is_retry=is_retry)
```

You're calling self._get_status() twice. Is it on purpose?

fernandolins (approved):
thx

PR #1352, opened 2026-07-30
[KNW-214] refactor: rename magic-code proto field to resend_cooldown_secs

fernandolins:
## Summary
Renames the magic-code response field `code_ttl_secs` to `resend_cooldown_secs` across the proto and the AnkiWeb signup/login UI. The old name described a code TTL, but the value the server returns is the per-email resend spacing (the seconds the client must wait before offering to resend). The new name matches the server contract in ankiweb.

- Rename `code_ttl_secs` → `resend_cooldown_secs` in `proto/user_backend/magic_code.proto` (signup and login responses).
- Wire `SignupCodeVerificationWidget` to seed its countdown from the server's `resend_cooldown_secs`, keeping a single internal `remaining_seconds` field for the timer (initial open, resend, and retry all share it).
- Update `_on_sign_up` and the retry call site to pass `remaining_seconds` consistently.
- Update unit tests to the new field/param names.

Pairs with the AnkiWeb PR that renames the field server-side and enforces the per-email resend cooldown.

## Related PR
https://github.com/ankitects/ankiweb/pull/30

## Test plan
- [ ] `pytest tests/addon/test_unit.py -k signup_code_verification`
- [ ] Manual: request a signup code, confirm the "Resend available in Ns" countdown starts from the server cooldown and the button re-enables at 0.
- [ ] Manual: fail a verification and confirm the retry screen resumes the remaining countdown instead of restarting.

### DIFF
--- a/ankihub/gui/ankiweb.py
+++ b/ankihub/gui/ankiweb.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 import sys
+import time
 from concurrent.futures import Future
 from enum import Enum
+from math import ceil
 from typing import Any, Callable, NoReturn, Union
 
 import aqt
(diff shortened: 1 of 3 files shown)


abdnh:
Cool.
I assume ResendCooldownTracker is not strictly required here with the server changes and is also easy to bypass by just restarting Anki (which is fine since the backend enforces the cooldown), but it's a nice UI improvement. I just suggest rewriting the docstring of ResendCooldownTracker since it can be misleading about this point.

Luc-Mcgrady (dismissed):

iamllama (dismissed):

abdnh (approved):

PR #1353, opened 2026-07-31
[KNW-177] Catch collection errors while media sync is active

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/KNW-177

<details>

<summary>Jira description for people with no access</summary>

Problem: The background sync does not cancel immediately when the user’s collection becomes unavailable in some cases (e.g. profile switched).

Solution: Review usage of mw.col in the sync logic and ensure there are checks to prevent errors.

</details>

## Proposed changes

Media sync runs in the background and relies on the existence of `aqt.mw.col` to get the media folder (which can be None when the profile is switched or a full sync is running). While the background task is canceled when the profile is switched, it takes time for cancellation to take effect and any access of `aqt.mw.col` there results in spurious errors. This is one of the most common error reports on [Discourse](https://community.ankihub.net/t/idk-this-is-annoying/596352)/[Sentry](https://ankihub.sentry.io/issues/6725784266/?project=6546414&query=TypeError&referrer=issue-stream&sort=recommended).

#1339 improved the situation by showing all kinds of errors in a dialog instead of throwing errors visibly at the user. This PR just adds a custom error message for that instead of showing TypeError/CollectionNotOpen exception messages.

## How to reproduce

1. Start a media sync (by downloading a large deck for example).
2. Open the media sync dialog from the toolbar button.
3. Switch the Anki profile quickly from _File>Switch Profile_ while the sync is starting (before the file count is displayed).
4. Depending on the timing, you should see "Collection is unavailable." displayed in the dialog.

## Screenshots and videos

image

### DIFF
--- a/ankihub/gui/media_sync.py
+++ b/ankihub/gui/media_sync.py
@@ -34,13 +34,19 @@ from aqt.qt import (
 )
 
 from .. import LOGGER
-from ..addon_ankihub_client import AddonAnkiHubClient
+from ..addon_ankihub_client import AddonAnkiHubClient, CollectionNotAvailableError, collection_or_error
 from ..ankihub_client.models import DeckMedia
 from ..common_utils import get_media_names_from_note_field, get_media_names_from_note_type
 from ..db import ankihub_db
 from ..settings import config, get_anki_profile_id
 from .operations import AddonQueryOp
-from .utils import error_icon, media_download_icon, media_sync_error_svg, media_sync_svg, media_upload_icon
+from .utils import (
+    error_icon,
+    media_download_icon,
+    media_sync_error_svg,
+    media_sync_svg,
+    media_upload_icon,
+)
 
 SHOW_MEDIA_PROGRESS_PYCMD = "ankihub_show_media_progress"
 TOOLBAR_BUTTON_ID = "ankihub_media_sync"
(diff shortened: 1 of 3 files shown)


tests/addon/test_integration.py
```diff
@@ -7965,6 +7960,41 @@ def test_download_media_with_filtering_by_referenced_media(
...
+                "get_deck_media_updates",
+                return_value=[
+                    DeckMediaUpdateChunk(media=[media], latest_update=latest_media_update),
+                ],
+            )
+            # Simulate closed collection
+            mocker.patch("aqt.mw.col", None)
+            with qtbot.captureExceptions() as exceptions:
+                media_sync.start_media_download()
+                qtbot.wait_until(lambda: media_sync._download_in_progress is False)
```
fernandolins, line 7999: This test is dedicated to downloading. 

Can you write a test for upload as well?

ankihub/gui/media_sync.py
```diff
@@ -111,7 +128,8 @@ def on_failure(exception: Exception) -> None:
             self._errors = [exception]
             self.refresh_sync_status()
             self._dialog.update_status(self._get_status())
-            raise exception
+            if not _is_collection_error(exception):
+                raise exception
```
fernandolins, line 132: Should we raise an exception with a message here?

``` 
raise Exception("Some specific message for this error.")
```

ankihub/gui/media_sync.py
```diff
@@ -146,7 +164,8 @@ def on_failure(exception: Exception) -> None:
...
```
fernandolins, line 168: same here

ankihub/gui/media_sync.py
abdnh: For collection errors? No because the goal of this PR is to hide this type of errors from the user. If we raise here, the error will show in ErrorDialog visibly and users will think something is broken.

fernandolins (approved):

PR #1362, opened 2026-08-06
[KNW-217] Fix AnkiHub sync failing silently after AnkiWeb errors

abdnh:
## Related issues

https://ankihub.atlassian.net/browse/KNW-217

## Proposed changes

AnkiHub sync fails silently without errors if the user is logged in to an unverified AnkiWeb account or if any other AnkiWeb issue occurs. We started to get more reports of this in Anki 26.08 from new users, as before that Anki used to incorrectly report that a full sync is required for completely new accounts, which made AnkiHub fail visibly with the “AnkiHub sync cancelled” message (fixed in [#5109](https://github.com/ankitects/anki/issues/5109)), reducing the possibility of users hitting the silent failure path.

The issue comes from Anki’s aqt.sync.get_sync_status() function ignoring connection/authentication errors (reason being that it’s intended for display of sync status in the toolbar). The solution implemented here involves using the underlying mw.col.sync_status directly and explicitly handling errors.

### Before

To reproduce the issue before this PR, create a new AnkiWeb account but do not verify it. Then log in to it in Anki and log in to an AnkiHub account and try to sync either using the Sync button or from the Deck Management screen. You should notice that nothing happens. If you check the console, you'll see the AnkiWeb error message as printed by Anki (look for for "sync status check failed").

### After

Any errors coming from AnkiWeb should now be visible.

### DIFF
--- a/ankihub/gui/operations/ankihub_sync.py
+++ b/ankihub/gui/operations/ankihub_sync.py
@@ -55,7 +55,10 @@ def get_sync_status(mw: aqt.main.AnkiQt, on_done: Callable[[Future], None]) -> N
 
     def wrapped_on_done(fut: Future) -> None:
         try:
-            fut.result()
+            status: SyncStatus = fut.result()
+            if new_endpoint := getattr(status, "new_endpoint", None):
+                if hasattr(mw.pm, "set_current_sync_url"):
+                    mw.pm.set_current_sync_url(new_endpoint)
             on_done(fut)
         except Exception as exc:
             LOGGER.warning("AnkiWeb status error", exc_info=exc)
--- a/ankihub/gui/operations/ankihub_sync.py
+++ b/ankihub/gui/operations/ankihub_sync.py
@@ -107,12 +107,7 @@ def _after_ankiweb_sync(on_done: Callable[[Future], None], skip_summary: bool) -
 
         _sync_with_ankihub_inner(on_done=on_done, skip_summary=skip_summary)
 
-    @pass_exceptions_to_on_done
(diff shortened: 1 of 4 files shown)


cjasm (dismissed):

ankihub/gui/operations/ankihub_sync.py
```diff
@@ -41,8 +47,31 @@ class _SyncState:
...
+        on_done(future_with_result(SyncStatus(required=SyncStatus.NO_CHANGES)))
+        return
+
+    def wrapped_on_done(fut: Future) -> None:
+        try:
+            fut.result()
+            on_done(fut)
+        except Exception as exc:
+            LOGGER.warning("AnkiWeb status error", exc_info=exc)
+            on_done(future_with_exception(AnkiWebSyncStatusError(exc)))
```
fernandolins, line 65: Here, after the `try... except`, the original `get_sync_status` from `aqt/sync.py` calls
``` python
        if out.new_endpoint:
            mw.pm.set_current_sync_url(out.new_endpoint)
```

For this new `get_sync_status`, is it unnecessary?

fernandolins (dismissed):
LGTM, I just left a comment, but I'm not sure if it is a relevant comment.

ankihub/gui/operations/ankihub_sync.py
abdnh: I was not sure if it's necessary since `aqt.sync.sync_collection()` sets it afterwards. I added it just in case.

fernandolins (approved):

PR #1364, opened 2026-08-07
[KNW-219] Fix Preferences AnkiHub auth so magic-code login and logout UI stay in sync

fernandolins:
# Related issues
[KNW-219](https://ankihub.atlassian.net/browse/KNW-219)

# Problem
Signing into AnkiHub from **Preferences → Third-party** did not notify the add-on, so feature flags were not refreshed and the AnkiWeb magic-code login dialog stayed off (old password dialog was shown). That flow could also reinstall the AnkiWeb add-on package.

Separately, after signing in from **AnkiHub menu → Sign In**, Preferences → Third-party still showed **Log in** instead of **Log out**, because Anki may reuse a cached Preferences window without refreshing the auth UI.

# How to reproduce

### Before
1. Open Preferences → Third-party → Log in to AnkiHub (native dialog)
2. Open Preferences → Syncing → Log in to AnkiWeb
3. Observe the old password-based AnkiWeb login dialog (magic-code dialog does not appear)
4. Sign out if needed, then sign in via AnkiHub menu → Sign In
5. Open Preferences → Third-party
6. Observe AnkiHub still shows **Log in** instead of **Log out**

### After
1. Open Preferences → Third-party → Log in to AnkiHub
2. Observe the add-on sign-in dialog opens
3. After login, open Preferences → Syncing → Log in to AnkiWeb
4. Observe the magic-code AnkiWeb login dialog (feature flag on)
5. Sign out, then sign in via AnkiHub menu → Sign In
6. Open Preferences → Third-party
7. Observe AnkiHub shows **Log out**

[KNW-219]: https://ankihub.atlassian.net/browse/KNW-219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/menu.py
+++ b/ankihub/gui/menu.py
@@ -240,10 +240,13 @@ class AnkiHubLogin(QWidget):
 def setup_preferences_ankihub_auth_patch() -> None:
     """Route Preferences → Third-party AnkiHub login/logout through the add-on.
 
-    Anki's native dialog always authenticates against production AnkiHub and then
-    installs the AnkiWeb add-on package. That breaks staging (invalid token → no
-    feature flags → magic-code AnkiWeb login patch stays off) and reinstalls a
-    duplicate add-on next to the development symlink.
+    Anki's native dialog does not go through Config.save_token, so feature flags are
+    not refreshed and the magic-code AnkiWeb login patch stays off. It also tries to
+    install the AnkiWeb add-on package, which can collide with a local install.
+
+    Also keep the Preferences AnkiHub login/logout buttons in sync when the user signs
+    in from the AnkiHub menu: Anki's DialogManager may reuse a cached Preferences
+    window, so we implement reopen() and refresh on token_change_hook.
     """
 
     def patched_ankihub_login(
@@ -268,11 +271,23 @@ def setup_preferences_ankihub_auth_patch() -> None:
         _sign_out_action()
         on_success()
 
+    def reopen_preferences(self: aqt.preferences.Preferences, *args, **kwargs) -> None:
+        # DialogManager reuses Preferences without recreating it; refresh auth UI.
+        self.update_login_status()
+
+    def refresh_preferences_ankihub_login_status() -> None:
+        prefs = aqt.dialogs._dialogs.get("Preferences", [None, None])[1]
+        if prefs is not None:
+            prefs.update_login_status()
+
     aqt.ankihub.ankihub_login = patched_ankihub_login
     aqt.ankihub.ankihub_logout = patched_ankihub_logout
     # preferences.py does `from aqt.ankihub import ankihub_login`, so patch both bindings.
     aqt.preferences.ankihub_login = patched_ankihub_login
     aqt.preferences.ankihub_logout = patched_ankihub_logout
+    aqt.preferences.Preferences.reopen = reopen_preferences  # type: ignore[method-assign]
+    if refresh_preferences_ankihub_login_status not in config.token_change_hook:
+        config.token_change_hook.append(refresh_preferences_ankihub_login_status)
     LOGGER.info("Set up Preferences AnkiHub auth patch.")
 
 
--- a/ankihub/gui/menu.py
+++ b/ankihub/gui/menu.py
@@ -6,7 +6,6 @@ from pathlib import Path
 from typing import Callable, Optional
 
 import aqt
-import aqt.ankihub
 import aqt.preferences
 from aqt import (
     AnkiApp,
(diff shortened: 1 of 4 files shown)


tests/addon/test_unit.py
```diff
@@ -2234,6 +2237,206 @@ def test_assignment_failure_when_patching_is_logged(self, mocker: MockerFixture)
         logger_mock.exception.assert_called_once()
 
 
+@pytest.mark.skipif(
+    not hasattr(aqt.profiles.ProfileManager, "set_ankihub_token"),
+    reason="ProfileManager.set_ankihub_token doesn't exist on this Anki version",
+)
```
Luc-Mcgrady, line 2243: Maybe this fix should be written into the code rather than as test exclusions in case someone uses that Anki version?

tests/addon/test_unit.py
```diff
@@ -2234,6 +2237,206 @@ def test_assignment_failure_when_patching_is_logged(self, mocker: MockerFixture)
...
+        original_flag = settings._native_ankihub_token_hook_installed
+        settings._native_ankihub_token_hook_installed = False
+        yield
+        ProfileManager.set_ankihub_token = original
+        settings._native_ankihub_token_hook_installed = original_flag
+
+    def test_set_ankihub_token_fires_token_change_hook(
+        self, mocker: MockerFixture, anki_session_with_addon_data: AnkiSession
+    ):
+        mocker.patch.object(aqt.mw.taskman, "run_on_main", side_effect=lambda fn: fn())
```
Luc-Mcgrady: This is what is causing the CI to fail. Remove this line from each of the tests and they run fine.

ankihub/gui/menu.py
```diff
@@ -225,10 +231,73 @@ def display_login(cls):
...
+        prefs = aqt.dialogs._dialogs.get("Preferences", [None, None])[1]
+        if prefs is not None:
+            prefs.update_login_status()
+
+    aqt.ankihub.ankihub_login = patched_ankihub_login
+    aqt.ankihub.ankihub_logout = patched_ankihub_logout
+    # preferences.py does `from aqt.ankihub import ankihub_login`, so patch both bindings.
+    aqt.preferences.ankihub_login = patched_ankihub_login
+    aqt.preferences.ankihub_logout = patched_ankihub_logout
+    aqt.preferences.Preferences.reopen = reopen_preferences  # type: ignore[method-assign, attr-defined]
```
abdnh: This will overwrite .reopen() if Anki ever defines it in the future.

abdnh (commented):
Monkey-patching does not feel great, and in my opinion, it's not something we should  encourage and resort to (unless absolutely necessary) now that we manage Anki and can make releases quickly.

What about simply hiding the AnkiHub section when the add-on is installed? It was only intended for onboarding new users quickly.

tests/addon/test_unit.py
fernandolins: Good point, but the safeguard already lives in `settings.py`: `setup_native_ankihub_token_hook()` checks `hasattr(ProfileManager, "set_ankihub_token")` and no-ops if missing. The skipif just reflects that there's nothing to call on that Anki version. I can add a test that simulates the missing attribute via monkeypatch.delattr if you'd like that covered explicitly.

ankihub/gui/menu.py
fernandolins: Good to know! Is there an alternative for it?
Maybe we can make this safer by using `anki.hooks.wrap(..., "around")` on reopen instead of a direct assignment. **Thoughts?**

That way we call through to whatever Anki's native reopen does

fernandolins:
> What about simply hiding the AnkiHub section when the add-on is installed? It was only intended for onboarding new users quickly.

@abdnh I understand your point and it is valid. But since we are only using it for Ankihub users, I think we can go as is.
Mainly because it was said that we shouldn't touch Anki for now.

However, when we are going to migrate this solution to Anki, I agree with you that this solution is not the best and we should think about alternatives.

And hiding it from preferences will depend on a conversation with the product team (Debs, Cani, Andrew, etc.) because it would be an important change.

ankihub/gui/menu.py
abdnh: That should work. We can check if `aqt.preferences.Preferences.reopen` is already defined and use `wrap()` in that case, and fall back to direct assignment otherwise.

fernandolins: done https://github.com/AnkiHubSoftware/ankihub_addon/pull/1364/commits/d7eca3b65a83654722f9f0aed9bda223de6e531e

abdnh (approved):

tests/addon/test_unit.py
Luc-Mcgrady: It might be a good idea. If there are no other tests to check that `setup_native_ankihub_token_hook` is functioning as intended.

PR #1365, opened 2026-08-10
[ACTV-445] Fix Step Deck tutorial browser/sidebar startup race crashes

cjasm:
## Related issues

- [x] Closes # [ACTV-445](https://ankihub.atlassian.net/browse/ACTV-445) [ACTV-446](https://ankihub.atlassian.net/browse/ACTV-446)

## Proposed changes
Fix Step Deck tutorial browser/sidebar startup race crashes

## Further comments
https://community.ankihub.net/tag/anki-tours/10068

[ACTV-445]: https://ankihub.atlassian.net/browse/ACTV-445?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[ACTV-446]: https://ankihub.atlassian.net/browse/ACTV-446?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/entry_point.py
+++ b/ankihub/entry_point.py
@@ -148,6 +148,8 @@ def _on_profile_did_open() -> None:
 
 
 def _on_profile_will_close() -> None:
+    if tutorial.active_tutorial:
+        tutorial.active_tutorial.end()
     media_sync.close_for_profile()
     LOGGER.info("Profile will close, stopping background threads.")
 
--- a/ankihub/gui/tutorial.py
+++ b/ankihub/gui/tutorial.py
@@ -1191,6 +1191,7 @@ class StepDeckTutorial(DeckBrowserOverviewBackdropMixin, Tutorial):
         self._deckoptions_saved = False
         self._browser: Optional[Browser] = None
         self._browser_closed_by_us = False
+        self._pending_browser_startup_completion = False
         self._unhook_browser: Callable[[], None]
 
     @ensure_mw_state("deckBrowser")
@@ -1202,6 +1203,7 @@ class StepDeckTutorial(DeckBrowserOverviewBackdropMixin, Tutorial):
         ).run_in_background()
 
     def end(self) -> None:
+        self._pending_browser_startup_completion = False
         self._unhook_browser()
         config.set_step_deck_tutorial_pending(False)
         return super().end()
@@ -1338,12 +1340,14 @@ class StepDeckTutorial(DeckBrowserOverviewBackdropMixin, Tutorial):
 
             browser = self._get_live_browser()
             if not browser:
-                LOGGER.exception("Skipping tutorial browser startup callback as browser is unavailable")
+                self._pending_browser_startup_completion = False
+                LOGGER.debug("Skipping tutorial browser startup callback as browser is unavailable")
                 return
 
             model = browser.sidebar.model()
             if not model:
-                LOGGER.exception("Skipping tutorial browser startup callback as sidebar model is unavailable")
+                self._pending_browser_startup_completion = False
+                LOGGER.debug("Skipping tutorial browser startup callback as sidebar model is unavailable")
                 return
 
             browser.sidebar.search_for(self._anking_deck_config.name)
@@ -1353,13 +1357,15 @@ class StepDeckTutorial(DeckBrowserOverviewBackdropMixin, Tutorial):
             except RuntimeError:
                 LOGGER.exception("Step Deck sidebar item unavailable during tutorial browser startup")
                 browser.sidebar.search_for("")
+                self._pending_browser_startup_completion = False
                 return
             search = aqt.mw.col.build_search_string(step_sidebar_item.search_node)
             browser.search_for(search)
             nonlocal is_startup
-            if is_startup:
+            if is_startup or self._pending_browser_startup_completion:
                 on_done()
             is_startup = False
+            self._pending_browser_startup_completion = False
 
         # There can be multiple sidebar refresh events at browser startup,
         # so we need to ensure we only call .next() once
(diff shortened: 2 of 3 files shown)


ankihub/gui/tutorial.py
```diff
@@ -1294,7 +1354,7 @@ def _build_deck_tree(*args: Any, **kwargs: Any) -> None:
             _old: Callable[..., None] = kwargs.pop("_old")
             args, kwargs, root = extract_argument(func=_old, args=args, kwargs=kwargs, arg_name="root")
             _old(*args, **kwargs, root=root)
-            aqt.mw.taskman.run_on_main(lambda: debouncer.schedule(self._browser, root))
+            aqt.mw.taskman.run_on_main(lambda: debouncer.schedule(self._get_live_browser() or aqt.mw, root))
```
abdnh, line 1385: Minor: With aqt.mw as the parent of the QTimer, the object will never get deleted. Not an issue now, but we can explicitly clean up just in case DebouncedDelayedCall is used for more frequent operations in the future:
```diff
diff --git a/ankihub/gui/tutorial.py b/ankihub/gui/tutorial.py
index 3bb6e854..5057c948 100644
--- a/ankihub/gui/tutorial.py
+++ b/ankihub/gui/tutorial.py
@@ -27,6 +27,7 @@ from aqt.qt import (
     QObject,
     QPoint,
     Qt,
+    QTimer,
     QToolButton,
     QVBoxLayout,
     QWidget,
@@ -171,6 +172,7 @@ class DebouncedDelayedCall:
         self._delay_ms = delay_ms
         self._pending_calls: List[Tuple[int, Tuple[Any], Dict[str, Any]]] = []
         self._next_sequence = 0
+        self._timer: Optional[QTimer] = None
 
     def schedule(self, parent: QObject, *args: Any, **kwargs: Any) -> None:
         self._next_sequence += 1
@@ -183,8 +185,11 @@ class DebouncedDelayedCall:
                 return
             self._pending_calls.remove((my_seq, args, kwargs))
             self._callback(*args, **kwargs)
+            if self._timer and not sip.isdeleted(self._timer):
+                self._timer.deleteLater()
+                self._timer = None
 
-        aqt.mw.progress.timer(self._delay_ms, run, repeat=False, parent=parent)
+        self._timer = aqt.mw.progress.timer(self._delay_ms, run, repeat=False, parent=parent)
 
```

ankihub/gui/tutorial.py
```diff
@@ -1276,11 +1319,28 @@ def hook_browser_startup(self, on_done: Callable[[], None]) -> Callable[[], None
...
+                LOGGER.debug("Skipping tutorial browser startup callback as sidebar model is unavailable")
+                return
+
+            browser.sidebar.search_for(self._anking_deck_config.name)
+            self._clear_sidebar_highlight(model.root)
+            try:
+                step_sidebar_item = self._find_step_deck_sidebar_item(root)
+            except RuntimeError:
+                LOGGER.debug("Skipping tutorial browser startup callback as Step Deck sidebar item is unavailable")
+                return
```
abdnh: An error here will leave the sidebar in a weird state where it's showing nothing with no search text. Can be fixed by adding `browser.sidebar.search_for("")` before returning.

abdnh (commented):
I'm concerned hiding errors will make it harder to debug similar issues in the future (especially since the cause is uncertain). If the tour breaks in a future Anki update or for any other reason, we'll only have an idea what's happening if users report that the tour disappears when the browser is open - we're less likely to get reports if there's no hard errors, which is why I originally made it explicitly raise an exception here.

While investigating what's causing this, I found that it also can be triggered by switching the Anki profile while the tour is open. The add-on does not handle this case, so the tour actually continues to show in the other profile and the error is triggered in the browser if the Step deck is not installed there. We should terminate the active tour in `_on_profile_will_close`.

ankihub/gui/tutorial.py
```diff
@@ -1276,15 +1335,37 @@ def hook_browser_startup(self, on_done: Callable[[], None]) -> Callable[[], None
...
+                self._pending_browser_startup_completion = False
+                LOGGER.debug("Skipping tutorial browser startup callback as sidebar model is unavailable")
+                return
+
+            browser.sidebar.search_for(self._anking_deck_config.name)
+            self._clear_sidebar_highlight(model.root)
+            try:
+                step_sidebar_item = self._find_step_deck_sidebar_item(root)
+            except RuntimeError:
+                LOGGER.exception("Step Deck sidebar item unavailable during tutorial browser startup")
```
abdnh, line 1364: The sidebar gets stuck in an infinite refresh loop if we don't call .end() here or `raise` the exception.

ankihub/entry_point.py
```diff
@@ -148,6 +148,8 @@ def _on_profile_did_open() -> None:
 
 
 def _on_profile_will_close() -> None:
+    if tutorial.active_tutorial:
+        tutorial.active_tutorial.end()
```
abdnh, line 152: This is not a new issue, but I noticed closing Anki while the tour is open no longer triggers it again on next startup. Is this intended?

ankihub/gui/tutorial.py
cjasm: solved on [f0e6add](https://github.com/AnkiHubSoftware/ankihub_addon/pull/1365/commits/f0e6addee07534d468ae10074e8b28e57e7ddcba)

ankihub/entry_point.py
cjasm: No, this is not intenteded, but it will be solved in another ticket

abdnh (approved):

PR #1368, opened 2026-08-12
[KNW-220] Fix signup-code retry so Get code stays available after cooldown ends

fernandolins:
## Related issue
KNW-220

## Problem
On the magic-code signup verification retry screen, failing to verify a code after the resend countdown had already ended restarted the full cooldown TTL.

That left **Get code** disabled again even though no new code had been requested, so users who mistyped an expired/invalid code could not immediately request another one.

### Before
1. Start signup with magic code and wait until the resend countdown reaches 0.
2. Enter an invalid or expired code and submit.
3. On the retry screen, **Get code** is disabled and a full “Resend available in …” countdown starts again — without the user clicking Get code / resend.

### After
1. Same steps as above.
2. On the retry screen, cooldown is `0`, **Get code** is enabled, and no countdown is shown.
3. Clicking **Get code** starts a new cooldown only after a successful request.
4. If verify fails while a countdown is still active, the remaining time is preserved and Get code stays blocked until it ends.

### DIFF
--- a/ankihub/gui/ankiweb.py
+++ b/ankihub/gui/ankiweb.py
@@ -979,10 +979,10 @@ class SignupCodeVerificationWidget(BaseSignupWidget):
                 tooltip("Sign-in successful!", parent=aqt.mw)
                 self._dialog._on_success()
             except Exception as exc:
-                if self._timer.remaining_seconds > 0:
-                    remaining_seconds = self._timer.remaining_seconds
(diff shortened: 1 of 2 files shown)


abdnh (commented):
LGTM.

Formatting check is failing.

abdnh (approved):

PR #1369, opened 2026-08-12
[KNW-221] Add resend cooldown and polish password-signup email verification UI

fernandolins:
## Related issue
KNW-221

## Problem
On the password-signup email verification screen, **Resend verification email** stayed enabled after being clicked, so users could fire multiple resend requests by accident or on purpose.

The screen also didn’t match the intended mockup (copy, Back/Sign in actions, countdown feedback), and multi-line instruction text could be clipped by Qt label/layout height handling.

### Before
image

1. Sign up with email + password.
2. On the verification screen, click **Resend verification email** repeatedly — each click can send another request; there’s no countdown feedback that a resend already happened.
3. The screen copy/actions don’t match the mockup (e.g. no Back-only footer, weaker “please change it” / instructions layout), and the last instruction line can be cut off.

### After
image

1. Sign up with email + password.
2. The verification screen starts with a **60s UI cooldown**; Resend is disabled and shows “Resend available in …s”.
3. After the countdown, Resend works once, then the 60s cooldown restarts. A second click while the request is in flight is blocked.
4. UI matches the mockup: verification copy with **please change it**, instructions bullets, countdown under the form, **Back** + **Sign in** (no Cancel), and instruction text is fully visible.

### DIFF
--- a/ankihub/gui/ankiweb.py
+++ b/ankihub/gui/ankiweb.py
@@ -912,8 +912,7 @@ class SignupEmailVerificationWidget(BaseSignupWidget):
                 f"Resend available in {remaining_secs}s" if remaining_secs else "Resend available."
             )
             self.status_label.setText(
-                f"If {self.email} account exists, we sent a message to its inbox.<br>"
-                + resend_available_status
+                f"If {self.email} account exists, we sent a message to its inbox.<br>" + resend_available_status
             )
             if not remaining_secs:
                 self.resend_button.setEnabled(True)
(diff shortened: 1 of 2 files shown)


ankihub/gui/ankiweb.py
```diff
@@ -45,12 +45,31 @@
 from .utils import error_icon, is_email
 
 EMAIL_INSTRUCTIONS = (
-    "Didn't receive an email?<ul><li>Check your spam folder. "
-    "Emails can end up there.</li><li>Resend the email when the countdown ends.</li></ul>"
+    "Didn't receive an e-mail?<ul><li>Check the spam folder. "
```
abdnh: Nitpick: "email" is more common and Breno agreed to use that (can't find the Figma discussion as the design was moved to a different Figma file).

abdnh (commented):
Looking good. Left a minor comment.

abdnh (approved):

PR #1374, opened 2026-08-17
[NRT-897] Keep personally protected fields when resetting local changes

cacoze:
## Related issues
- [x] [NRT-897](https://ankihub.atlassian.net/browse/NRT-897)
- Cloned from [TRIAGE-320](https://ankihub.atlassian.net/browse/TRIAGE-320)

## Proposed changes
Reset local changes was stripping `AnkiHub_Protect::*` tags before reimporting, so auto-protected edits (and the protect tag) were discarded. CS confirmed that protected fields should not be affected by Reset — same as a normal sync.

This only changes the two browser call sites (`Reset local changes` on notes, and reset-all-local-changes for a deck) to `strip_personal_protect_tags=False`. The strip flag stays so we can restore the old Reset behavior later without rewriting the reset path. No UI or copy changes.

## How to reproduce
1. Sync an AnkiHub deck. Confirm **Automatically protect fields when edited** is on.
2. Edit a shared field, blur so auto-protect adds `AnkiHub_Protect::<Field>`.
3. Right-click the note → **Reset local changes**.
4. The edit and the protect tag should still be there.

Also check:
- Unprotect a field, edit it, Reset → that field reverts; remaining protected fields stay.
- Edit a globally protected field (e.g. Lecture Notes) → Reset leaves it.
- Delete a note, Reset the deck → the note comes back.

To actually revert a protected edit: remove the protect tag, then Reset (or use Anki undo).

## Screenshots and videos
N/A — no UI changes.

## Further comments
NRT-769 made Reset strip personal protect tags so auto-protect wouldn't make Reset a no-op. CS uses Reset as a routine support step and confirmed protected fields should be left alone. Unprotected fields still reset.

[NRT-897]: https://ankihub.atlassian.net/browse/NRT-897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TRIAGE-320]: https://ankihub.atlassian.net/browse/TRIAGE-320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/browser/browser.py
+++ b/ankihub/gui/browser/browser.py
@@ -510,7 +510,7 @@ def _on_reset_local_changes_action(browser: Browser, nids: Sequence[NoteId]) -> 
         tooltip("Reset local changes for selected notes.", parent=browser)
 
     aqt.mw.taskman.with_progress(
-        task=lambda: reset_local_changes_to_notes(filtered_nids, ah_did=ankihub_did),
+        task=lambda: reset_local_changes_to_notes(filtered_nids, ah_did=ankihub_did, strip_personal_protect_tags=False),
         on_done=on_done,
         label="Resetting local changes...",
         parent=browser,
@@ -597,7 +597,7 @@ def _on_reset_deck_action(browser: Browser):
         tooltip(f"Reset local changes to deck <b>{deck_config.name}</b>")
 
     aqt.mw.taskman.with_progress(
-        lambda: reset_local_changes_to_notes(nids, ah_did=ah_did),
+        lambda: reset_local_changes_to_notes(nids, ah_did=ah_did, strip_personal_protect_tags=False),
         on_done=on_done,
         label="Resetting local changes...",
         parent=browser,
--- a/ankihub/gui/browser/browser.py
+++ b/ankihub/gui/browser/browser.py
@@ -510,7 +510,7 @@ def _on_reset_local_changes_action(browser: Browser, nids: Sequence[NoteId]) -> 
         tooltip("Reset local changes for selected notes.", parent=browser)
 
     aqt.mw.taskman.with_progress(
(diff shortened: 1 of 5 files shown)


ddevdan (changes requested):
2 findings 
- a (this PR) — no test pins False at either browser call site, though the db-check wiring is pinned (test_integration.py:6811-6829)
 - b (this PR) — test_reset_local_changes_to_notes:6724-6779 still passes True and asserts the old outcome

cacoze:
Addressed both findings:

- **a** — `test_browser_reset_does_not_strip_personal_protect_tags` now pins `strip_personal_protect_tags=False` on both browser Reset entry points (selected notes and reset-all-to-a-deck).
- **b** — `test_reset_local_changes_to_notes` now calls with `False` and asserts the auto-protected Front edit (and its protect tag) survive.

abdnh (dismissed):

cacoze:
Addressed the latest review:

- **True coverage** — `test_reset_local_changes_to_notes_strips_personal_protect_tags` (and `..._strips_protect_all_tag`) call with `True` again, including that `AnkiHub_Protect::All` is not preserved.
- **Duplicate False tests** — renamed to `test_reset_local_changes_to_notes_keeps_personally_protected_fields`. The main test now covers personal Front + global-only Back; the renamed one covers unprotected Front vs personal-only Back.
- **Global protection** — main test no longer puts `AnkiHub_Protect::Back` on the note, so the surviving Back edit is from `protected_fields` only.
- **Protect::All** — `test_reset_local_changes_to_notes_keeps_fields_when_protect_all_tag` (`False`) and the strip variant (`True`).

abdnh (approved):

ddevdan (approved):

PR #1375, opened 2026-08-18
[NRT-903] Claim the trial-ended message when showing the upsell

ddevdan:
## Related issues
- Closes [NRT-903](https://ankihub.atlassian.net/browse/NRT-903)
- Server side: AnkiHubSoftware/ankihub#3910 — removes the field from user details and adds the claim endpoint. **Must be deployed before this PR is released.**

## Proposed changes
- Add `AnkiHubClient.claim_trial_ended_message()` for `POST /users/me/trial-ended-message/claim`.
- Claim the message when the upsell is displayed and pick the trial-ended title from the result. Restores copy that became unreachable once the server stopped sending `show_trial_ended_message` in user details.
- Fail soft: a failed claim logs a warning and shows the generic copy. Needs an explicit `.failure()` handler because `AddonQueryOp` re-raises by default.
- Report `upgrade_cta_viewed` / `upgrade_cta_clicked` with `source` and `surface` matching the web taxonomy. The native dialog never loads a webview, so the add-on CTA was previously invisible to analytics.

## How to reproduce
1. Sign in as a user without Smart Search access whose trial has ended.
2. Open a deck and click the Smart Search button.
3. The upsell shows "Your Trial Has Ended! 🎓✨"; reopening shows the generic title.

## Further comments
Depends on the server endpoint, which is implemented but not yet in production — ship server first. Until then the fail-soft path shows the generic copy.

The `tests/addon` suite was not run locally (segfaults on macOS); CI is the first check on those three tests.

[NRT-903]: https://ankihub.atlassian.net/browse/NRT-903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/gui/flashcard_selector_dialog.py
+++ b/ankihub/gui/flashcard_selector_dialog.py
@@ -70,11 +70,10 @@ UPSELL_SOURCE_TRIAL_ENDED = "trial_ended"
 UPSELL_SOURCE_GENERIC = "generic_upsell"
 
 # Third value in the web's `surface` taxonomy, alongside `web_app` (browser) and
-# `anki_webview` (webview hosted in Anki), which are set in _analytics_state.html:17.
-# Note that Smart Search's webview is also served inside the add-on and reports
-# `anki_webview`, so this value means specifically the add-on's native Qt layer.
-# Both surfaces open the same plans page with no marker of their own, so this property
-# is the only thing preventing double-counting.
+# `anki_webview` (webview hosted in Anki). Smart Search's webview is also served inside
+# the add-on and reports `anki_webview`, so this value means specifically the add-on's
+# native Qt layer. Both surfaces open the same plans page with no marker of their own,
+# so this property is the only thing preventing double-counting.
 UPSELL_SURFACE = "anki_addon"
 
 
@@ -95,8 +94,8 @@ def _track_upsell_event(event_name: str, source: str) -> None:
                 distinct_id=user_id,
                 event_name=event_name,
                 properties={
-                    # Key and values mirror the web upsell's taxonomy so the two channels can
-                    # be reconciled later (ModalUpsellContent.html:40,47 emits `source`).
+                    # `source` and `surface` mirror the web upsell's taxonomy so the two
+                    # channels can be reconciled later.
                     "source": source,
                     "surface": UPSELL_SURFACE,
                     "user": user_id,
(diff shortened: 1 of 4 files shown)


ankihub/gui/flashcard_selector_dialog.py
```diff
@@ -83,6 +139,36 @@ def on_button_clicked(button_index: int) -> None:
         default_button_idx=1,
         callback=on_button_clicked,
     )
+    _track_upsell_event("upgrade_cta_viewed", source)
+
+
+def _show_upsell(user_details: dict, parent=aqt.mw) -> None:
```
cacoze: `user_details` is unused now. I'd drop it from the signature and change the call site to `lambda _: _show_upsell(parent)`.

Keeping the argument makes it look like the copy still comes from `GET /users/me`, which is the thing this PR is stopping.

ankihub/gui/flashcard_selector_dialog.py
```diff
@@ -59,17 +66,66 @@ def _get_non_embed_url(self) -> str:
...
 
 
-def _show_upsell(user_details: dict, parent=aqt.mw) -> None:
-    show_trial_ended_message = user_details.get("show_trial_ended_message", False)
+UPSELL_SOURCE_TRIAL_ENDED = "trial_ended"
+UPSELL_SOURCE_GENERIC = "generic_upsell"
+
+# Third value in the web's `surface` taxonomy, alongside `web_app` (browser) and
+# `anki_webview` (webview hosted in Anki), which are set in _analytics_state.html:17.
+# Note that Smart Search's webview is also served inside the add-on and reports
```
cacoze: The `anki_addon` vs `web_app` / `anki_webview` distinction is worth keeping — that's the contract.

The pointer at `_analytics_state.html:17` (and `ModalUpsellContent.html:40,47` a few lines below) lives in the Django repo, so the path and line numbers will rot and aren't greppable from here. Naming the properties (`source`, `surface`) is enough.

cacoze (commented):
Two nits from reading this next to the server PR. Neither is a merge blocker — the claim-at-display-time path looks right.

ankihub/gui/flashcard_selector_dialog.py
ddevdan: Fixed both

cacoze:
Safe to merge as https://github.com/AnkiHubSoftware/ankihub/pull/3910 was already deployed

cacoze (approved):

PR #1379, opened 2026-08-28
[KNW-225] Do not report media sync errors caused by stale DB entries

abdnh:
## Related issues

[KNW-225](https://ankihub.atlassian.net/browse/KNW-225)

## Proposed changes

The media sync progress dialog implemented in #1339 uncovered an issue where the add-on keeps trying to download media files that were deleted from the backend ([TRIAGE-321](https://ankihub.atlassian.net/browse/TRIAGE-321)). As long as the backend objects are not modified, the add-on will keep reporting errors with such files on every sync.

## How to reproduce

This should not be normally reproducible with fresh deck installs. It usually happens for old users as they are more likely to have media entries belonging to deleted files in their local AnkiHub database. See report: https://community.ankihub.net/t/unable-to-download-media-file-deck-assets-e77aedfe-a636-40e2-8169-2fce2673187e-tmplvqol9-png-403-forbidden/607281

[KNW-225]: https://ankihub.atlassian.net/browse/KNW-225?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TRIAGE-321]: https://ankihub.atlassian.net/browse/TRIAGE-321?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ankihub/ankihub_client/ankihub_client.py
+++ b/ankihub/ankihub_client/ankihub_client.py
@@ -681,7 +681,9 @@ class AnkiHubClient(AnkiWebClientMixin):
                 media_remote_path=media_remote_path,
                 status_code=response.status_code,
             )
-            raise AnkiHubMediaDownloadError(response, media_remote_path)
+            # Ignore missing files (e.g. stale media entries in the database)
+            if response.status_code != 403:
+                raise AnkiHubMediaDownloadError(response, media_remote_path)
 
     def stop_background_threads(self) -> None:
         """Can be called to stop all background threads started by this client."""
--- a/tests/client/test_client.py
+++ b/tests/client/test_client.py
@@ -1474,10 +1474,13 @@ class TestDownloadMedia:
 
             assert len(futures) == len(media_names)
 
+    @pytest.mark.parametrize("status_code, raises", [(404, True), (403, False)])
     def test_with_failed_download(
         self,
         requests_mock: Mocker,
         next_deterministic_uuid: Callable[[], uuid.UUID],
+        status_code: int,
+        raises: bool,
     ):
         with tempfile.TemporaryDirectory() as temp_dir:
             client = AnkiHubClient(local_media_dir_path_cb=lambda: Path(temp_dir))
@@ -1485,7 +1488,7 @@ class TestDownloadMedia:
             deck_id = next_deterministic_uuid()
             requests_mock.get(
                 f"{DEFAULT_S3_BUCKET_URL}/deck_assets/{deck_id}/missing.png",
-                status_code=404,
+                status_code=status_code,
                 reason="Not Found",
             )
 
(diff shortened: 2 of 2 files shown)


ankihub/ankihub_client/ankihub_client.py
```diff
@@ -681,7 +681,9 @@ def _download_media(self, media_file_path: Path, media_remote_path: str) -> None
                 media_remote_path=media_remote_path,
                 status_code=response.status_code,
             )
-            raise AnkiHubMediaDownloadError(response, media_remote_path)
+            # Ignore missing files (e.g. stale media entries in the database)
+            if response.status_code != 403:
```
fernandolins, line 685: I didn't get it. Did you mean that every 403 means there is no file?

abdnh: S3 returns 403 if the file is missing for unauthenticated users, unless if bucket is modified to grant the `ListBucket` permission (in which case it returns 404 instead). So there's no guaranteed way to tell if the file is actually missing or it's an actual access permission problem in the bucket (which will be a big problem and completely break deck installs so practically 403 only happens when the file is missing).

This is a stopgap solution until a proper solution that stops the requests from being made in the first place is implemented ([TRIAGE-321](https://ankihub.atlassian.net/browse/TRIAGE-321)).

fernandolins: Thanks for the clarification. I agree this can be a reasonable stopgap.

However, as it stands, whether a download succeeds or fails (403) makes no difference. How do we notify the UI or register the download result? We should differentiate, right?

Furthermore, can we identify why it gave a 403? Even if it is done at another time? Because if the database says that the file `exists_on_s3=True`, can we confirm that it still exists? And if it doesn't exist, should we change the value in the database?

abdnh: > However, as it stands, whether a download succeeds or fails (403) makes no difference. How do we notify the UI or register the download result? We should differentiate, right?

My thinking was that it's fine not to differentiate them in the UI for now since we know they only add noise and increase CS tickets, as no used media files should be actually missing. It's only that the add-on is keeping record of deleted files. The errors are still logged though so we can still inspect them if needed.

> Furthermore, can we identify why it gave a 403?

Not from the add-on. The solution described in [TRIAGE-321](https://ankihub.atlassian.net/browse/TRIAGE-321) (running a backend script that marks `DeckMedia` objects with `exists_on_s3=False` as modified) will cause the add-on to automatically update its local database on the next sync, which will help with most cases of the issue. But thinking of this more, having the add-on reports these errors visibly will help us identify `DeckMedia` objects that should be updated in the backend, so maybe it's better to hold off on hiding the errors for now - we got a few reports so far and I already updated a few hundreds of entries in the backend based on the reports so we should get less reports now.

Closing this.

